Issue #13, opened 2016-12-21
Test code needed


karllark:
Setup test code in the astropy test format.  Existing test code could be used, but not sure.
Setup the continuous testing with travis as part of this.

PR #93, opened 2017-03-18
Travis ci setup

karllark:
Travis CI working!  
Closes issue #13.


beast/physicsmodel/helpers/hdfstore.py
```diff
@@ -1,27 +1,30 @@
...
-            hd['/subdir/table'] = d
-            hd['/subdir1/array'] = c
-            #check values
-            print hd.keys()
-            print hd['subdir/table']
-            print hd['subdir1/array']
-            hd.remove_node('/subdir1', recursive=True)
-
 """
+# this code block fails when 'python setup.py test' is run
```
mfouesneau: Switch to a code block so it still appears in the API doc.

beast/physicsmodel/stars/ezmist/simpletable.py
```diff
@@ -14,24 +14,27 @@
...
-    >>> t.set_alias('logL', 'logLLo')
-    # make a query on one or multiple column
-    >>> q = s.selectWhere('logT logL', '(J > 2) & (10 ** logT > 5000)')
-    # q is also a table object
-    >>> q.plot('logT', 'logL', ',')
-    # makes a simple plot
-    >>> s.write('newtable.fits')
-    # export the initial subtable to a new file
 """
+
```
mfouesneau: Same. I guess the >>> are telling the test to run it. I suppose the imports are missing or what?

beast/physicsmodel/helpers/hdfstore.py
karllark: This is the error that I get with this code.  Basically, it is running the code on the >>> lines and expecting the lines afterwards to be the output.   As this is not happening, it fails.  I thought the best solution would be to comment it out for now and come back and "fix" it.  Not quite sure what the fix is yet.  Better to get the travis ci working right now so that we have at least basic testing of the code.  Lots of documentation fixing needed anyway.

_________________________________________ [doctest] beast.physicsmodel.helpers.hdfstore __________________________________________
001 """ HDFStore
002 High level interface to PyTables for reading and writing structures to disk
003 Could work either as a (key, value) storage or as a classic tables.file.File object.
004     >>> import numpy as np
Expected:
        #make a store
        with HDFStore('tmp.hd5', mode='w') as hd:
            #make some variables
            d = {}
            d['a'] = np.arange(10, dtype=float)
            d['b'] = np.arange(10, dtype='int')
            c = np.random.normal(0, 1, (10, 10))
            d['c'] = c
            #put values into the store
            hd['/subdir/table'] = d
            hd['/subdir1/array'] = c
            #check values
            print hd.keys()
            print hd['subdir/table']
            print hd['subdir1/array']
            hd.remove_node('/subdir1', recursive=True)
Got nothing

beast/physicsmodel/helpers/hdfstore.py:4: DocTestFailure

beast/tools/helpers.py
```diff
@@ -355,12 +357,11 @@ def val_in_unit(varname, value, defaultunit):
...
     >>> print(val_in_unit('r', r, 'degree'))
-    # UserWarning: Variable r does not have explicit units. Assuming `degree`
-    <Quantity(0.5, 'degree')>
+    0.5 degree
 
     >>> r = 0.5 * unit['degree']
     >>> print(val_in_unit('r', r, 'degree'))
-    <Quantity(0.5, 'degree')>
+    0.5 degree
     """
```
karllark: Note the changes here.  I got this to work.  
Basically, the >>> lines are run and the result in on the next line.  I had to update as the result has to be exactly what is returned by the print statement.

karllark: More information: https://docs.python.org/2.7/library/doctest.html

beast/physicsmodel/helpers/hdfstore.py
karllark: Found a solution.  Will go through and fix.  If we just want an example in the documentation, but not to have doctest see if it produces correct results, use:

> Example::
>     import numpy as np
>     #make a store
>     with HDFStore('tmp.hd5', mode='w') as hd:
>         #make some variables
>         d = {}
>         d['a'] = np.arange(10, dtype=float)
>         d['b'] = np.arange(10, dtype='int')
>         c = np.random.normal(0, 1, (10, 10))
>         d['c'] = c
>         #put values into the store
>         hd['/subdir/table'] = d
>         hd['/subdir1/array'] = c
>         #check values
>         print hd.keys()
>         print hd['subdir/table']
>         print hd['subdir1/array']
>         hd.remove_node('/subdir1', recursive=True)

beast/tools/helpers.py
karllark: We can add python 3 tests once the beast code runs in python 3.  Can't do that yet.

karllark:
The "example::" format works with sphinx.  I have tested it.

PR #102, opened 2017-06-16
Code Corrections: SMC ExtCurve, Isochrone Min logAge

lcjohnso:
Two Edits:

1) SMC extinction curve (`Gordon03_SMCBar()`) -- and therefore `Gordon16_RvFALaw()` suffered from an implementation error concerning the application of C_4*F(x) term of the extinction curve.  When `draine_extend=False`, code was applying term to full UV portion of curve at lam < 2700 Ang (x > 3.7 um^-1) rather than only at x > 5.9 um^-1.  This error is caused by the `ind` index selection not being properly redefined for the `else` block. This plot to shows the effect of bug to the SMC curve, and the corrected version:
edit_smccurvecorrection

2) Minor edits to enable logA values down to 5.0 as allowed by MIST isochrones.

### DIFF
--- a/beast/physicsmodel/dust/extinction.py
+++ b/beast/physicsmodel/dust/extinction.py
@@ -406,9 +406,10 @@ class Gordon03_SMCBar(ExtinctionLaw):
                      ( ((xspluv) ** 2 - (x0 ** 2)) ** 2 + (gamma ** 2) *
                        ((xspluv) ** 2 ))
             
-            # FUV portion  
+        # FUV portion  
+        ind = np.where(x >= 5.9)
+        if np.size(ind) > 0:
             if draine_extend:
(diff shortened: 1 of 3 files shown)


karllark:
This looks good to me.  Good catch.
The failing travis-ci check is ok.  It is on the development version of astropy.  Probably should not be testing this with python 2.7.

galaxyumi:
Thanks for catching this, Cliff! I think this was from my local version of BEAST, which has been using the extinction curves (i.e., Draine curves) below the LyC regime all the time. So I did not worry about the else block. But this correction looks correct to me.

karllark:
Checked with regress_check.py and it looks good.  This change did not effect the phat_small test case.  Likely due to the wavelength of this change being at too short of a wavelength.

PR #103, opened 2017-06-16
Noise Model: shift to percentile-based calculation

lcjohnso:
I propose that the noise model calculation use a percentile-based method to estimate sigma instead of using the standard deviation of (Output-Input) values for the `toothpick` noise model.  This change was motivated by unrealistically large estimates of sigma obtained with the standard deviation, driven by non-gaussian outliers that created long low-amplitude tails to the (Output-Input) distribution of AST results.

In this pull request, I substitute a new `_compute_percentile_bins` function for the original `_compute_stddev_bins` function, where sigma = (p84-p16)/2.

Here are plots compare the old (top) and new (bottom) results for the noise model (red lines show median +/- 1 sigma envelopes) versus F475W AST results (blue), where y-axis shows Delta(mag) and x-axis shows F475W input flux:
noisemod-old_475
noisemod-new_475

Optional Future Additions:
1) Add keywords to relevant functions enabling a choice between the two methods.  Currently, the new percentile-based method simply replaces the stddev method for the toothpick model.
2) Implement more complex noise model that accounts for skew in (out-in) distribution.

### DIFF
--- a/beast/observationmodel/noisemodel/toothpick.py
+++ b/beast/observationmodel/noisemodel/toothpick.py
@@ -93,26 +93,17 @@ class MultiFilterASTs(NoiseModel):
             print(e)
             print('Warning: Mapping failed. This could lead to wrong results')
 
-    def _compute_stddev_bins(self, magflux_in, magflux_out, nbins=30,
-                             min_per_bin=5,
-                             completeness_mag_cut=80, name_prefix=None,
-                             asarray=False):
-        """
-        Alias of _compute_stddev_bins
-        """
-        return self._compute_stddev_bins(magflux_in, magflux_out,
-                                    nbins=nbins, min_per_bin=min_per_bin,
-                                    completeness_mag_cut=completeness_mag_cut,
-                                    name_prefix=name_prefix,
-                                    asarray=asarray)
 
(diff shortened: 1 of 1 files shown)


karllark:
I like this change and am in favor.

Looking at the code, it looks like most of `_compute_percentile_bins` is a copy of `_compute_stddev_bins.`  With only a few lines changed.  It would be much easier if the `_computer_stddev_bins` just had a keyword that affected those few lines instead of having lots of duplicate code.  The name of  the function could change to be more general if that would help.   Fewer lines of code to maintain then.
 
So, this is a request to have a single function to compute the 'stddev' values with an option to use the percentiles or the standard stddev equation.

lcjohnso:
Yes, the new function was just a clone of the stddev-based function.  I'm happy to combine into single function and push the update.

What should the default behavior and function name be?  Due to the known weakness of stddev, I propose primary functionality be percentile-based, provide stddev functionality as a keyword-accessible feature, and rename function more generally: `_compute_sigma_bins`.  How does that sound?

karllark:
This sounds great.  I like all the suggestions.

lcjohnso:
Requested edits complete!  Also included: edit docstrings, removal of old func alias.

Issue #44, opened 2017-02-22
Adjustable A(V), R(V), and f_A priors


karllark:
The A(V), R(V), and f_A priors are currently hard coded. They are:

A(V): flat
R(V) - f_A: triangular shape based on Gordon16 model - see BEAST paper
Better would be to have the form and necessary variables set in the datamodel.py file or something else. Something else may be better for eventual use by the megabeast.

Two motivations.

Provide flexibility for the input of the dust priors for BEAST runs
Provide flexibility for the eventual use of the BEAST dust physicsmodel by the MegaBEAST

karllark:
I am starting on this now.  Just a FYI.

Issue #43, opened 2017-02-22
Adjustable age, mass, and metallicity priors


karllark:
The age, mass, and metallicity priors are currently hard coded.  They are:
- age: constant star formation rate (linear age)
- mass: a Kroupa IMF
- metallicity: flat

Better would be to have the form and necessary variables set in the datamodel.py file or something else.  Something else may be better for eventual use by the megabeast.

Two motivations. 
- Provide flexibility for the input of the stellar priors for BEAST runs
- Provide flexibility for the eventual use of the BEAST stellar physicsmodel by the MegaBEAST

karllark:
Working on this during the BEAST HackDay.

Conversations have lead to the following plan.
Mass: pick an IMF, provide the parameters (e.g., Salpeter, Kroupa)
Age: n bins in log(age)
Metallicity: (more discussion) use metallicity grid and impose some kind of smoothness as a function of age would be good.  Will investigate.

PR #104, opened 2017-06-20
Adjust dust prior

karllark:
Full adjustable priors for the dust parameters A(V), R(V), and f_A.   This addresses issue #44.

There are 3 possible models - flat, lognormal, and two_lognormal.
Which model is used is controlled by variable in datamodel.py.

For example,
`av_prior_model = {'name': 'flat'}`
gives a flat prior between the min/max of the input A(V) range.
and 
`av_prior_model = {'name': 'lognormal', 'max_pos': 2.0, 'sigma': 1.0, 'N': 10.}`
gives a lognormal prior with a maximum at A(V) = 2.0 a width of 1.0 and a peak value of 10.  The peak value is not important for the BEAST, but will be used for the megaBEAST.

All three dust parameters can have priors.  Setting them all to 'flat' results in the previously hard-coded behavior.

These changes have been checked with the regress_check.py code.

In addition, minor updates in the long-term effort to make the code work in python 3.
Also, travis-ci testing updated to include the stable (not dev) version of astropy and newer versions of numpy.  This solves the failure for the astropy dev version.


karllark:
Yes, I think we should be adding them to the BEAST.  If it is useful for the megaBEAST, it is also useful for the BEAST.  I think that if I were to find a good model shape with the megaBEAST, I could potentially use this in the BEAST as a way to "improve" the BEAST fits and saved files (lnp, etc.).

karllark:
I made it specific to dust for now.  This was mainly as the priors for the stellar side are applied when the spectral grid is created, not the sed grid.  And there already is code for the stellar priors.  I do agree that it could(should?) be expanded to include the stellar priors.  Probably good to do this so that if we ever come up with stellar+dust joint priors, it will be much easier to implement.  But I think this should be done in a separate branch/pull request to keep things clear.  There is an issue for this (issue #43) and I will work on this later.  Right now I am focusing on getting the megaBEAST running with adjustable dust priors.  Will add adjustable stellar priors after it is working (should be pretty easy given how I will be modifying the existing megabeast code).

beast/physicsmodel/creategrid.py
```diff
@@ -387,6 +406,10 @@ def make_extinguished_grid(spec_grid, filter_names, extLaw, avs, rvs,
         if npts == 0:
             raise AttributeError('No valid points')
     else:
+        dustpriors = PriorWeightsDust(avs, av_prior_model,
+                                      rvs, rv_prior_model,
+                                      [1.0], {'name': 'flat'})
```
lcjohnso: Replace `{'name': 'flat'}` here with `fA_prior_model` which is set to that dictionary value as default keyword.

beast/physicsmodel/models.py
```diff
@@ -274,12 +290,20 @@ def make_seds(outname, specgrid, filters, av=[0., 5, 0.1], rv=[0., 5, 0.2],
     if fA is not None:
         fAs = np.arange(fA[0], fA[1] + 0.5 * fA[2], fA[2])
         g = creategrid.make_extinguished_grid(specgrid, filters, extLaw,
-                                              avs, rvs, fAs,
+                                              avs, 
+                                              rvs, 
+                                              fAs, 
+                                              av_prior_model,
+                                              rv_prior_model,
+                                              fA_prior_model,
```
lcjohnso: Pass the prior models as keywords: `av_prior_model=av_prior_model`, etc.

beast/physicsmodel/models.py
```diff
@@ -274,12 +290,20 @@ def make_seds(outname, specgrid, filters, av=[0., 5, 0.1], rv=[0., 5, 0.2],
...
                 add_spectral_properties_kwargs=add_spectral_properties_kwargs,
                                               absflux_cov=absflux_cov)
     else:
         g = creategrid.make_extinguished_grid(specgrid, filters, extLaw,
-                                              avs, rvs,
+                                              avs, 
+                                              rvs, 
+                                              av_prior_model,
+                                              rv_prior_model,
```
lcjohnso: Same as above -- pass the prior models as keywords: `av_prior_model=av_prior_model`, etc.

This is necessary in the `else` block, otherwise av_prior_model is passed as the 6th param (here, fAs).

beast/physicsmodel/creategrid.py
karllark: Done.

beast/physicsmodel/models.py
karllark: Done.

beast/physicsmodel/models.py
karllark: Good point.  Done.

lcjohnso:
One oddity with using the lognorm prior for Av: it always assigns weight=0. at Av=0. -- effectively eliminating a grid point.  I'm not sure this is necessarily bad, but something to be aware of.

heddyarab:
In the MegaBEAST, we deal with that by adopting 2 lognormals, the first one having a peak at Av=0.

karllark:
How is it possible to have a lognormal with a peak at A(V)=0.  Since this is a lognormal, it is undefined when x=0.  Or am I missing something?
And I would have thought for the megabeast we would want the first one to be near zero, but not at zero.  The foreground dust should have a peak/mean above zero as it is due to dust in the MW.

heddyarab:
You're right Karl, the first component in MegaBEAST is Av~0.

Issue #71, opened 2017-02-25
Documentation for setting up datamodel.py


karllark:
Documentation for how to setup datamodel.py for running the BEAST is needed.  Example is useful, but more information in documentation form would be best.

lcjohnso:
Added initial documentation via PR https://github.com/BEAST-Fitting/beast/pull/124, but this could be expanded.

Anyone (especially new users) have suggestions for valuable / helpful items to add?

karllark:
Need to do a thorough review and see if we have enough docs on this topic now.

karllark:
Closing this issue.  We have what we have.  In the new work for v2.0, we'll have to update and that will be the chance to make sure we've got enough.

PR #124, opened 2017-10-13
Adjust Av and Age priors, allow verify_params failures

lcjohnso:
Here I enable three optional features:
1) Add an exponential Av prior model.
2) Allow the removal of the "constant in linear age" prior to prefer the grid's intrinsic logarithmic spacing.
3) Allow verify_params failures to print to screen, but continue without exiting.  This is helpful when using non-fA dust curves and fixed R_V values.

While not included in this PR, I would be happy to add info about what needs to be changed in run_beast.py and datamodel.py to use these updates, but I'll wait until the datamodel.py documentation page is drafted (Issue https://github.com/BEAST-Fitting/beast/issues/71).

### DIFF
--- a/beast/physicsmodel/prior_weights_dust.py
+++ b/beast/physicsmodel/prior_weights_dust.py
@@ -85,6 +85,23 @@ def _two_lognorm(xs,
     normalization = np.trapz(pointwise, x=xs)
     return pointwise / normalization
 
+def _exponential(x, a=2.0, N=1.):
+    """
+    Exponential distribution
+    Parameters
+    ----------
+    x: vector
+       x values
+    a: float
+       Single parameter in exp: a*e^-ax
+    N: floats
+       Multiplicative factor
+    Returns
+    -------
+    exponential computed on the x grid
+    """
+    return N * np.exp(-1. * a * x)
+
 class PriorWeightsDust():
     """
     Compute the priors as weights given the input grid
@@ -195,6 +212,7 @@ class PriorWeightsDust():
           flat = flat prior on linear A(V)
           lognormal = lognormal prior on linear A(V)
           two_lognormal = two lognormal prior on linear A(V)
+          exponential = exponential prior on linear A(V)
         """
         if model['name'] == 'flat':
             self.av_priors = np.full(len(self.av_vals),1.0)
@@ -211,6 +229,10 @@ class PriorWeightsDust():
                                           sigma2=model['sigma2'],
                                           N1=model['N1'],
                                           N2=model['N2'])
+        elif model['name'] == 'exponential':
+            self.av_priors = _exponential(self.av_vals,
+                                          a=model['a'],
+                                          N=model['N'])
         else:
             print('**error in setting the A(V) dust prior weights!**')
             print('**model ' + model['name'] + ' not supported**')
--- a/beast/physicsmodel/prior_weights_dust.py
+++ b/beast/physicsmodel/prior_weights_dust.py
@@ -93,8 +93,9 @@ def _exponential(x, a=2.0, N=1.):
     x: vector
        x values
     a: float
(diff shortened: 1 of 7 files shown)


karllark:
In the verify_params changes, I'm not sure having the exit being overridden for all possible exits is a good idea.  In fact, ideally the override would not be done at all, but the verify_params logic changed to handle the two cases mentioned - no fA and/or fixed Rv extinction curves.  @lcjohnso: Is such a change in the logic not possible/straightforward?

lcjohnso:
Re: @karllark's suggestion, I added logic to verify_params that allows for non-`exit()` warnings in the case that:

1) `fAs = None`, as needed for case where non-Gordon16 extinction curve is used.

2) Parameter grids are defined as single values (e.g., defining sequence is [0.0, 0.0, 0.1] leading to single grid point at 0.0).

I'm leaving the `noexit` option implemented (could still have use in future), but its use is discouraged as it might allow non-physical grid/parameter values.

lcjohnso:
Also: I added initial documentation of datamodel.py/run_beast.py, including HowTo instructions for implementing new optional features.

This addition is a good step towards closing Issue https://github.com/BEAST-Fitting/beast/issues/71 -- please comment there on additions that might still be needed before closing issue.

beast/physicsmodel/prior_weights_dust.py
```diff
@@ -85,6 +85,23 @@ def _two_lognorm(xs,
...
 
+def _exponential(x, a=2.0, N=1.):
+    """
+    Exponential distribution
+    Parameters
+    ----------
+    x: vector
+       x values
+    a: float
+       Single parameter in exp: a*e^-ax
```
karllark: This should be N*e^-ax.

beast/physicsmodel/prior_weights_dust.py
```diff
@@ -85,6 +85,23 @@ def _two_lognorm(xs,
...
+    N: floats
```
karllark: Is this parameter needed?  Shouldn't this be set by the desire to have this prior integrate to 1 over the min/max age range?  If so, then it is not needed and should be set inside the function.

beast/tools/verify_params.py
```diff
@@ -13,39 +13,43 @@
...
-        exit()
+        if not noexit: exit()
 
     if (param_max-param_min) < param_step:
-        print((param_name + ' step value greater than (max-min)'))
-        exit()
-
-def verify_one_input_format(param, param_name, param_format, param_lim):
+        if param_max-param_min == 0.:
+            print('Warning: '+param_name+' grid is single-valued.')
```
karllark: Nit picky point.  I don't think this is a warning.  There is no problem with single value parameters in a grid.  I worry that having it issue a warning makes it sound like this is bad.  Maybe something less "scary" like "Info"?

docs/beast_setup.rst
```diff
@@ -0,0 +1,73 @@
+Running the BEAST
```
karllark: Great to have this doc file!  Many thanks for starting it.  I know there is material I will add once I think about it more.

beast/tools/verify_params.py
lcjohnso: Changed to "Note:"

beast/physicsmodel/prior_weights_dust.py
lcjohnso: I was following the format of the other prior choices.  Will leave as is.

beast/physicsmodel/prior_weights_dust.py
lcjohnso: I corrected the notation, but am leaving the param as "a" -- could use "decay_rate" or "decay_param" but that seems overkill.  Instead, I added additional text in docstring.

Issue #116, opened 2017-08-18
Linux chokes on special character in MIST json file


meredith-durbin:
I just tried running the BEAST on Linux, and received the following traceback:

```python
[mdurbin@kix phat_small]$ python run_beast.py -potf
Traceback (most recent call last):
  File "run_beast.py", line 22, in <module>
    from beast.physicsmodel.model_grid import (make_iso_table,
  File "/astro/users/mdurbin/run_beast/beast/beast/examples/phat_small/beast/physicsmodel/model_grid.py", line 9, in <module>
    from . import creategrid
  File "/astro/users/mdurbin/run_beast/beast/beast/examples/phat_small/beast/physicsmodel/creategrid.py", line 27, in <module>
    from .stars import isochrone
  File "/astro/users/mdurbin/run_beast/beast/beast/examples/phat_small/beast/physicsmodel/stars/isochrone.py", line 22, in <module>
    from .ezmist import mist
  File "/astro/users/mdurbin/run_beast/beast/beast/examples/phat_small/beast/physicsmodel/stars/ezmist/__init__.py", line 1, in <module>
    from .mist import get_one_isochrone, get_t_isochrones
  File "/astro/users/mdurbin/run_beast/beast/beast/examples/phat_small/beast/physicsmodel/stars/ezmist/mist.py", line 38, in <module>
    _cfg = json.load(f)
  File "/astro/apps6/opt/anaconda/envs/py3/lib/python3.5/json/__init__.py", line 265, in load
    return loads(fp.read(),
  File "/astro/apps6/opt/anaconda/envs/py3/lib/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 884: ordinal not in range(128)
```

It looks like the ascii codec can't handle the umlaut in "Strömgren" in beast/physicsmodel/stars/ezmist/mist.json. This is likely a Linux quirk, as I haven't seen it happen on macOS. I got around it by just removing the umlaut, but a more robust fix would be useful eventually.

Details:
CentOS Linux release 7.3.1611 (Core) 
Python 3.5.2

mfouesneau:
It's part of the description of the filter set. Get rid of the special character without any consequence

karllark:
Fixed with pull request #127.

PR #127, opened 2017-10-18
Fix ezmist special character issue

lcjohnso:
Fixes Issue https://github.com/BEAST-Fitting/beast/issues/116: removes troublesome umlaut from JSON file.


karllark:
Great!

Issue #132, opened 2017-10-28
Error in run_beast -a


benw1:
With the latest version, I cannot seem to generate AST lists because of an error in quantity.py in astropy.  Has anyone else seen this?

(astroconda) MacBook-Pro-2:14675_LMC-9441ne-25634 ben$ ./run_beast.py -a
Auto-detected type: fits
Auto-detected type: fits
/Users/ben/Cpipelines/metal_2/data/14675_LMC-9441ne-25634/beast/fitting/fit_metrics/__init__.py:72: UserWarning: C-code cannot be imported, did you run make? Using python code instead
  warnings.warn('C-code cannot be imported, did you run make? Using python code instead')
Auto-detected type: hd5
Auto-detected type: fits
Auto-detected type: fits
Traceback (most recent call last):
  File "./run_beast.py", line 154, in <module>
    pick_positions(filename,separation)
  File "/Users/ben/Cpipelines/metal_2/data/14675_LMC-9441ne-25634/beast/observationmodel/ast/make_ast_xy_list.py", line 39, in pick_positions
    catalog = datamodel.get_obscat(datamodel.obsfile,datamodel.distanceModulus,datamodel.filters)
  File "/Users/ben/Cpipelines/metal_2/data/14675_LMC-9441ne-25634/datamodel.py", line 290, in get_obscat
    obs = GenFluxCatalog(obsfile, filters=filters)
  File "/Users/ben/Cpipelines/metal_2/data/14675_LMC-9441ne-25634/datamodel.py", line 213, in __init__
    self.setFilters( filters )
  File "/Users/ben/Cpipelines/metal_2/data/14675_LMC-9441ne-25634/datamodel.py", line 268, in setFilters
    _, vega_flux, _ = v.getFlux(filters)
  File "/Users/ben/Cpipelines/metal_2/data/14675_LMC-9441ne-25634/beast/observationmodel/vega.py", line 55, in getFlux
    for k in filters ])
  File "/Users/ben/miniconda2/envs/astroconda/lib/python2.7/site-packages/astropy/units/quantity.py", line 1142, in __iter__
    .format(cls=self.__class__.__name__))
TypeError: 'Quantity' object with a scalar value is not iterable
(astroconda) MacBook-Pro-2:14675_LMC-9441ne-25634 ben$

meredith-durbin:
[...]

PR #134, opened 2017-10-28
Fix minor bugs in make_ast_xy_list.py

meredith-durbin:
The deprecated distance modulus parameter was still lurking in make_ast_xy_list (#132), so here I remove that, and also fix some column access issues and delete an unused pylab import.


karllark:
Looks reasonable to me.  @benw1 : can you test?

meredith-durbin:
@benw1 The old way of indexing appears to only work in python 2. The new way works with both 2 and 3.

karllark:
@benw1 : is it ok to merge this pull request into the master?

benw1:
Yes.  Meredith explained the index order differences, so it is ready to go.

Thanks,
Ben

PR #138, opened 2017-11-18
Typo correction and minor fix to support megabeast work

karllark:


karllark:
Reviewed at BEAST telecon 20 Nov 2017.

Issue #130, opened 2017-10-24
Documentation needed for filter response functions


karllark:
At a minimum, documentation is needed for how to add new filter response functions to the BEAST filter database.  
More complete documentation giving the existing filters with their names would be even better.

mfouesneau:
Beast adapted pyphot, which has a complete documentation (or as complete as possible)
http://mfouesneau.github.io/docs/pyphot/

karllark:
Would be useful to have the BEAST documentation make plots showing the filter response functions in the filter database.  This can be done as part of the sphinx documentation run (regen plots everytime!).

mfouesneau:
You have ~200 filters, plus the ones defined in some examples by Yumi et al.
Maybe just showing a list and how to plot the response?
(http://mfouesneau.github.io/docs/pyphot/libcontent.html)

lea-hagen:
I'm going to start working on this one.

galaxyumi:
I also need to add a code called "extra_filters.py" (I think we need a better name for it) that does generate user defined filters (e.g., top hat or any transmission filters that are not in our filter library). I used it for computing number of LyC photons.

lea-hagen:
Does this code already exist in a useful form?  We're doing things as part of the hack day today, so I'd be glad to start with what you've already written.

galaxyumi:
I just sent a code to Karl asking him to add it to the BEAST. I used the code with python 2.7 version BEAST. But I think it would work with python 3 version as well.

karllark:
Still need documentation for how to add a new filter to the database.  Thus, leaving this issue open.

Issue #115, opened 2017-08-17
"extra_filters" script seems to be missing


meredith-durbin:
In `beast/examples/phat_small/run_beast.py`, there is the following commented line:

`from extra_filters import make_integration_filter, make_top_hat_filter`

I can't find anything named `extra_filters` anywhere (including in `old`), and the `make_integration_filter` and `make_top_hat_filter` functions haven't been moved elsewhere or reimplemented under other names as far as I can tell. I'm working from branch py3_compat and have also checked master.

It'd be good to have this or similar functionality reinstated for working with filters that aren't in the database. (Or, if I've just missed where `extra_filters` lives, I can update the commented line.)

galaxyumi:
Karl, did you remove it or move it to a more proper directory? 
That is a necessary code to get UV flux or photon flux within a user defined wavelength range.

karllark:
Yumi, I never put it in the example directory.  It should be added to the tools directory.  I can do this or you can.  :-)

PR #148, opened 2017-11-27
Adding a file called "extra_filters.py" in the observationmodel directory

galaxyumi:
#115 #130 This code generates user defined filters.

### DIFF
--- a/beast/observationmodel/extra_filters.py
+++ b/beast/observationmodel/extra_filters.py
@@ -1,5 +1,5 @@
 import numpy as np
-from beast.core.phot import IntegrationFilter
+from phot import IntegrationFilter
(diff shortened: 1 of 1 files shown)


beast/observationmodel/extra_filters.py
```diff
@@ -0,0 +1,140 @@
+import numpy as np
+from beast.core.phot import IntegrationFilter
```
karllark: This is with the old directory structures.
I think this should be 
> from phot import IntegrationFilter

karllark:
Looks good.  Would be good to add documentation and tests.  But as this is a statement for most of the code, I'll go ahead and merge.  :-)

PR #150, opened 2017-11-27
added documentation about filters

lea-hagen:
I added a few lines describing how to access information in filters.hd5, and a long table with the list of filters currently included.  This may need to be organized differently.


karllark:
Addresses issue #130.

Issue #167, opened 2018-01-22
add BEAST slack space to documentation


lea-hagen:
@karllark says there's a BEAST slack space.  It should go into the documentation.

PR #168, opened 2018-01-24
Adding in slack, test changes, and commit video documentation

karllark:
Will address issue #167.

### DIFF
--- a/docs/beast_development.rst
+++ b/docs/beast_development.rst
@@ -188,7 +188,7 @@ in a web browser by point to files in the `docs/docs/_build/html subdirectory`.
 
     .. code:: shell
 
-      $ python setup.py test build_docs
+      $ python setup.py build_docs
 
 Collaborating and Contributing
 ==============================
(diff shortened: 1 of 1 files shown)


karllark:
Can you post the errors text you are getting when running the test code?
Please tell me what version of python and numpy you are using.  If the tests did not complete, then you will not see the htmlconv directory giving the coverage results.

For the documentation, it is in the docs/_build/html directory.

lea-hagen:
Turns out I was trying to run it on a mac instead of linux, so the first tests are fine, and I can see the html files in the htmlconv directory.  But I haven't been able to figure out how to get the html for the docs to appear.  After running `python setup.py test build_docs`, it finishes with a green line that says `17 passed, 7 skipped`, but there is no _build in the docs directory.

karllark:
Good catch.  The command should be 'python setup.py build_docs'.  I've updated the doc file to have the correct command.
I was wondering if the tests failed because of being run on a Mac.  I've been thinking we might need to have a way to alter the tests based on the computer used.  Either by having a different set of files to download or by allowing for small differences.  I have opened #169 for this issue.

lea-hagen:
That solved it!  Looks good.

PR #172, opened 2018-02-09
Fix: mass_prior_weights uninitialized

drvdputt:
I was playing around a bit and came across this bug.

mass_prior_weights is not initialized in the 'else' section of the 'if' block. It is then used in the for loop below it. If the 'else' section is entered the first time the 'if' block is evaluated, this leaves mass_prior_weights uninitialized, crashing the program with an error.

Fix suggested: Set to np.zeros(1), just like mass_grid_weights. Would this produce correct behaviour?


karllark:
Good catch.

Issue #128, opened 2017-10-18
Structured array assignment in numpy 1.13


meredith-durbin:
As of numpy version 1.12.1, this warning is raised:

> beast/beast/external/eztables/backends/hdf5backend.py:135: FutureWarning: Assignment between structured arrays with different field names will change in numpy 1.13.
> 
> Previously fields in the dst would be set to the value of the identically-named field in the src. In numpy 1.13 fields will instead be assigned 'by position': The Nth field of the dst will be set to the Nth field of the src array.
> 
> See the release notes for details
>   t.append( tab.data.astype(t.description._v_dtype) )

This may pose a problem for writing the grid HDF5 files in future; for now I'm just putting it here as an early heads up.

karllark:
@mfouesneau : Can you comment on this issue?  Is there something that needs to be changed to fix this?

mfouesneau:
that probably means the order of the data will change but could also means breaking changes.

We'll have to wait for numpy 1.13 tests.

karllark:
This has now happened.  See https://travis-ci.org/BEAST-Fitting/beast/jobs/330410262

mfouesneau:
Can you debug and see if the issue is on the t.append or the tab.data?

karllark:
I will check later.  Thanks for the info of where to look.

karllark:
Fixed!

PR #173, opened 2018-02-15
Put columns in correct order before structured numpy array assignment.

drvdputt:
See #128


karllark:
Looks great.  The travis ci test with python 3.6 is now passing and was not previously.  Most excellent!

karllark:
Waiting for all the checks to complete before merging.

Issue #155, opened 2017-11-30
PIP installable version of the BEAST


karllark:
For those that are only interested in running the BEAST, we should have a version installable via PIP.

Things needed for a good build and PIP install are:

- [x] location of library files outside of source code (and a way to know this in the code)
- [x] alternative method for downloading library files (other than manual)
- [x] example runs outside the source code (separate repository?)
- [x] Documentation of all the above

karllark:
See discussion in the closed issue #78 for possible locations of the library files.

karllark:
Potentially useful.   Of particular interest is how to install command-line tools.

http://python-packaging.readthedocs.io/

karllark:
Working on this.

PR #174, opened 2018-02-23
Pip installable version

karllark:
Update the beast so it can be installed via pip.  Locations for library files other than in the src directory now possible (see docs).
Useful in general and needed for megabeast work (testing, docs, etc.).

Example beast run files moved to new repository in BEAST-Fitting.
https://github.com/BEAST-Fitting/beast-examples

Addresses issue #155.

Using atom editor, so some PEP8 changes for extra characters automatically done.


karllark:
Probably need to redo this pull request.  The updates for astropy-helpers as well as other pull requests adding information to the examples directory point to this.  The examples directory was removed in this pull request as the examples were moved to a separate repository.

drvdputt:
If we want to have to keep the run script in the main repo, while moving the datamodel.py examples somewhere else, we need to have another way of loading the latter.

Right now this is done with 'import datamodel', but this means that a copy of the run script is needed in the directory where the relevant datamodel.py is located.

If we want to keep using a python file for configuration, there are a couple of ways to import a python file/module from a file path directly. 
https://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path

This seems to be the safest way to me
```
import os
import importlib.util

datamodelfile = os.path.join(os.getcwd(), 'datamodel.py')

spec = importlib.util.spec_from_file_location("datamodel", datamodelfile)
datamodel = importlib.util.module_from_spec(spec)
spec.loader.exec_module(datamodel)

print(datamodel.project)
```
The other options is of course to use a more limited, text based configuration file.

karllark:
Too many conflicts.  New pull request against current master.

PR #175, opened 2018-02-23
Add optional redshift parameter

meredith-durbin:
This pull request implements redshift as an optional parameter in `datamodel.py`. The wavelength array in `make_spectral_grid` is simply multiplied by (1 + redshift), with a default redshift of 0. I've tested it on both current examples.

If there is a better way to implement this, or if keeping the rest frame wavelengths somewhere in the spectral grid is important, please let me know!

### DIFF
--- a/beast/examples/metal_production/datamodel.py
+++ b/beast/examples/metal_production/datamodel.py
@@ -127,6 +127,9 @@ absflux_a_matrix = absflux_covmat.hst_frac_matrix(filters)
 # distance modulus to the galaxy
 distanceModulus = 18.5 * units.mag
 
+# redshift of galaxy
+redshift = 0.000875 # LMC redshift from SIMBAD
+
 ################
 
 ### Stellar grid definition
--- a/beast/examples/metal_production/run_beast_production.py
+++ b/beast/examples/metal_production/run_beast_production.py
@@ -110,12 +110,18 @@ if __name__ == '__main__':
         else:
             extra_kwargs = None
 
+        if hasattr(datamodel, 'redshift'):
+            redshift = datamodel.redshift
+        else:
+            redshift = 0
+
         # generate the spectral library (no dust extinction)
         (spec_fname, g_spec) = make_spectral_grid(
             datamodel.project,
             oiso,
             osl=datamodel.osl,
             distance=distance,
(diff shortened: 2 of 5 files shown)


galaxyumi:
I am a bit confused. Is this related with adding distance parameter in the model grid?

meredith-durbin:
No, currently this is totally independent of distance; my thinking was that for most use cases the object of inquiry would be nearby enough that peculiar velocity dominates. I could certainly add a way to get redshift from distance if that would be useful though.

karllark:
We should keep the redshift to be just the known redshift of a galaxy.  For the kind of galaxies that the BEAST will be applied to - nearby and well resolved - the redshift is not well correlated with the distance (peculiar motions and all that).  Calling this redshift may be a bit confusing.  It might be less confusing to call it the galaxy velocity.

galaxyumi:
Then why do we need redshift here? I might be missing something. Can you specify the relevant issue number?

meredith-durbin:
Whoops, I guess I should have opened an issue beforehand. JD realized that redshift/peculiar velocity/whatever you want to call it might be important for some of the narrowband work we've been doing. For the vast majority of use cases it won't have any noticeable impact, but it was fairly trivial to add in.

Karl, I can rename the redshift parameter to galaxy_velocity or something similar if that would be helpful.

karllark:
@meredith-durbin : Yes, please rename.  I think galaxy_velocity would be clearer to users what it is.  Also, having default units of km/s would be good.  When I think about nearby galaxy velocities, I think of them in km/sec, not redshift.

karllark:
And please use astropy.units for the value of the speed of light when calculating the wavelength shift (e.g., 1 + v/c).

karllark:
Also, a little documentation needed in beast_setup.rst.

meredith-durbin:
I've changed things in `datamodel.py` from redshift to velocity in km/s, and the red/blueshift is now calculated in `run_beast.py` using the speed of light from astropy constants. Documentation to follow.

karllark:
I resolved the conflicts using the edit on github option.  But this did not quite fix everything as the distances are now applied it a subroutine and the redshift passing needs to be added to this subroutine.

Issue #38, opened 2017-02-21
Add distance to the BEAST


karllark:
Adding distance is needed for work in the Magellanic Clouds.

The fastest way to do this from human standpoint is to do multiple BEAST runs with different distances on a uniform grid.  The results can be used to regenerate all the standard outputs of the BEAST with the addition of the distance information included *and* distance as the 7th fit parameter.  

Basically (given a uniform distance grid):

- For all the 1D pPDFs (except distance), the 1D pPDFs for the the different distances can be added together to generate the 1D pPDFs including marginalization over distance.
   **need to make sure we are not normalizaing the 1D pPDFs for output**
- For the the 1D pPDF for distance, this can be created by simply summing one of the 1D pPDFs for any other parameter for each distance as this provides the marginalization over all other parameters
- For the stats file, all the parameters p50, p13, p87, exp, etc. need to be regenerated from the 1D pPDFS (just like they are done during the fitting)
- For the max values, use the set of Pmax values for each distance and find the maximum one to set a the new Pmax value (and min chisqr, etc.)
- For the sparse nD likelihoods, I think the solution is to merge all of them into one file, created a sparse likelihood that is n times larger than for a single distance run (more thinking about if this is correct from the sampling standpoint)

Code is needed to do all this semi-seamlessly to avoid human error (e.g., include distance grid in datamodel.py and provide scripts to setup the n distance BEAST grids and merge the results).

karllark:
One thought: we should put in distance just like the other 6 parameters.  Thus, we can make a grid that has a range of distances if we want, or have a grid with one distance.  This would provide flexibility in putting in distance.  Maybe the grid will get small enough to allow enough distance bin points - or computers will get enough RAM - or the BEAST will get setup for true parallel computations in the sense of different parts of the grid to difference nodes.

drvdputt:
I think that adding distances to the grid should not increase the amount of RAM used actually. Since the distance just rescales everything, it wouldn't make sense to store separate SEDs for the models at different distances.
When the distances are put into the model grid as an extra axis, we would need some trick to refer to the same SED for multiple grid points.
I think the better option is to keep the distances on a separate grid, and implement a loop somewhere near the end, when the probabilities are calculated. I will investigate these sections of the code and see how this fits in.

mfouesneau:
Distance can be first a deterministically optimized parameter for each model on the grid. Then you get a posterior predictive distribution of distances. 

optimal distance:
[...]

PR #178, opened 2018-02-23
Add Distance as 7th fitting variable

drvdputt:
Addresses issue #38.


drvdputt:
In its current state, this branch passes the tests on my (linux) machine. The Travis tests show some weird errors with Cython though. It is unclear to me where these are coming from.

I tried a few different versions of numpy and python, but was not able to reproduce these errors.

karllark:
The travis issues should be solved once the merge has happened.  The version of astropy-helpers needed updating.  I've now gotten onto the list for getting automated pull requests for this.  And just merged such a pull request.

karllark:
Need some documentation.  Please look in the docs subdirectory and see if it is obvious where to put such documentation.  If not, ask and I will look and provide suggestions.

karllark:
One complication that we need to solve is that all the examples will be moved to a separate repository (already setup in BEAST-Fitting) when the pip_installable pull request is merged.  So, conflicts!  Maybe I should just go ahead and do the merge this (and other pull requests) and then  redo the pip_installable pull request.  But maybe someone else has a clever idea for this.

drvdputt:
So, the basic functionality for the distances should be almost there. I just have a couple of
questions/comments to make.

- general

  - There are some formatting changes in the files I modified, mostly whitespace trimming. You can    hide the latter by adding '?w=1' to the end of the github URL when viewing a diff.

- examples

  - In the data model files, the distances are given as a range (or a single number) and unit (length or mag).

  - In the run files, these values are passed to one of the functions, where the distances are calculated and applied.

  - There is a new example, which is the same as phat_small but with multiple distances. It is called phat_small_multidistance, and has its own datamodel file.

  - I put a way of loading the datamodel as a python model directly from a given file path in run_beast.py for phat_small. This way, I can run the multidistance example without having to copy phat_small/run_beast.py and phat_small/data/ (just use symlinks).

    QUESTION:
    It seems kind of hacky though, and the way of doing this has changed a couple of times (see stackoverflow link in source code), so I'll need some feedback on whether this is acceptable.

    It's hardly a necessary feature, especially since we're planning to move the examples somewhere else. But it does give the first step towards having a general run script.

- grid creation

  - In creategrid.py, there is now an apply_distance_grid function

  - This function is used in model_grid.py: make_spectral_grid().

    This function now has 'distance' and 'distance_unit' as arguments. 'distance' can be a float or [min, max, step]. distance_unit can be either units.mag or any length unit, and defaults to parsec.

    This way, I've ensured that using a single number in parsec (not a 1-element list) also works, so the tests for this function still work without modification.

    The distance grid is then created linearly in distance modulus space or directly in distance space depending on the unit given.

- plotting

  - I've modified the plot_indiv_fit script, so that it also displays the distance (even when there's only one distance)

    QUESTION:
    Is it acceptable that the distance is always added to the grid (and plotted), even if there is only one?

- documentation
  
  - 'distances' and 'distance_unit' are mentioned in the explanation about the datamodel.

karllark:
All the changes look good.
The tests are failing now on a different issue.  Has to do with docstrings that have example code.  Just need to comment these out (e.g., ">>>" need to be removed).  The tests of the docstrings are trying to run the code with ">>>" and it is not working.

karllark:
Probably should fix the failing tests in a separate pull request.  Then all new pull requests will be tested against the fixed code.  Safer that way.  I can have a look at fixing this issue in the next week or so.

karllark:
Or someone else can.  :-)

karllark:
Tests have been fixed by a separate pull request.  Hence, tests now passing - yeah!

Issue #181, opened 2018-02-26
Numpy 1.15 will return view instead of copy


drvdputt:
Seems like there will be another change to some of the numpy code related to #128. 

I get a FutureWarning with numpy 1.14. 

> Traceback (most recent call last):
>   File "run_beast.py", line 125, in <module>
>     add_spectral_properties_kwargs=extra_kwargs)
>   File "/run/user/1090/beast-work/beast/physicsmodel/model_grid.py", line 369, in make_extinguished_sed_grid
>     gk.writeHDF(seds_fname, append=True)
>   File "/run/user/1090/beast-work/beast/physicsmodel/helpers/gridbackends.py", line 347, in writeHDF
>     self.grid.write(fname, tablename='grid', append=True)
>   File "/run/user/1090/beast-work/beast/external/eztables/table.py", line 188, in write
>     return manager().write(self, filename, **kwargs)
>   File "/run/user/1090/beast-work/beast/external/eztables/backends/hdf5backend.py", line 136, in write
>     data_in_correct_order = tab.data[list(datatype_wanted.names)]
>   File "/run/user/1090/Software/miniconda3/lib/python3.6/site-packages/numpy/core/records.py", line 513, in __getitem__
>     return obj.view(dtype=(self.dtype.type, obj.dtype))
> FutureWarning: Numpy has detected that you may be viewing or writing to an array returned by selecting multiple fields in a structured array. 
> 
> This code may break in numpy 1.15 because this will return a view instead of a copy -- see release notes for details.
> 

I don't think this will break anything, since we aren't writing anything to the array anyway.

karllark:
Is there a fix possible?

mfouesneau:
```python 
data_in_correct_order = tab.data[list(datatype_wanted.names)]
data_as_ndarray = data_in_correct_order.astype(datatype_wanted)
```
replaced by 

```python
data_as_ndarray = np.array([tab.data[k] for k in datatype_wanted.names], 
                                               dtype=datatype_wanted)
```

PR #183, opened 2018-02-28
Work around FutureWarning.

drvdputt:
First create a zeros array of the correct data type, then explicitly copy the columns, to avoid the warning mentioned in #181


karllark:
Great!

Issue #184, opened 2018-03-16
Tools to pick stars for ASTs based on background density map


drvdputt:
For modelling IC63, we decided that the different intensities of the background (gas emission) would play an important role in the AST parameter space.

I recently made a script that measures the background for each source, and then divides an image into spatial bins which each have a certain background density.

This background density map can then be used to evenly spread out the artificial stars across these regions. The spatial bins can for example be grouped into background density bins, and then every group of spatial bins gets a number of artificial starts assigned to it.

While this and the distance fitting features of #178 are related (they will both be used for IC63), I will be putting these scripts in a separate branch + pull request.

drvdputt:
Note that this also relates to #24

PR #185, opened 2018-03-21
AST input list sampling based on background intensity

drvdputt:
#184 
The two main contributions here are:

- A script (in the tools directory) which creates a map of the background intensity based on a fits image. It also needs a GST catalog to help with masking the point sources. Some constants, such as the size of the point source masks, are hardcoded in this script (i.e. it works for the IC63 Hubble images).

- A new function in make_ast_input_list which uses the map produced by the above script, to evenly sample regions of similar background intensity.

The exact output format of the AST input file might not be correct yet, but the machinery is in place now.

Note that there's also some changes to the examples, as I've removed the datamodel import from make_ast_input_list.


karllark:
Looks good.

Issue #187, opened 2018-04-04
Standalone run script with absolute datamodel import, avoid other datamodel imports


drvdputt:
It is possible to import a python file as a module directly from a file path, although the exact way of doing it is different between python versions.
https://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path

By importing datamodel from the working directory this way, we would be able to generalize the run script, put it in a scripts directory, and install it for the pip version. Users then no longer need to copy the `phat_small/run_beast` example. Having to put a beast symlink might also be avoided this way, as the run script will be in a more convenient location, with direct access to the beast module.

Note that there are still some other places (e.g. `make_ast_xy_list`) the code where datamodel is imported, usually out of convenience. This usually happens out of convenience, for passing a bunch of filenames for example. We should get rid of these imports, and pass whatever was used as function parameters instead.

karllark:
Now being tracked in issue #320.  This will be more general and give the same capabilities.

PR #188, opened 2018-04-04
WIP: First step: Make run_beast.py runnable from another directory

drvdputt:
Some example code showing how a module can be imported from an absolute
path, in this case the current working directory.

There is another way commented out which uses runpy, which returns the datamodel namespace as a dictionary.

This PR can be considered merge-ready once it has addressed all the issues mentioned in #187.


karllark:
Going to close this pull request.  I think we will go a different route in the future.  Very much in favor of what is happening for the megabeast in this context (BEAST-fitting/megabeast#22).

Issue #95, opened 2017-03-20
Requiring AST file when it is not needed


benw1:
I need to generate a model grid in order to generate a beast-friendly list of fake stars, but when I run run_beast.py -p, I get an error:

(iraf27) BenjamiiamssMBP:14675_LMC-5665ne-12232 ben$ ./run_beast.py -p &
[1] 13624
(iraf27) BenjamiiamssMBP:14675_LMC-5665ne-12232 ben$ Auto-detected
type: fits
Auto-detected type: fits
/Users/ben/Cpipelines/14675_LMC-5665ne-12232/beast/fitting/fit_metrics/__init__.py:72:
UserWarning: C-code cannot be imported, did you run make? Using python code instead
  warnings.warn('C-code cannot be imported, did you run make? Using
python code instead')
astfile does not exist. Please provide the file path.

Why is it looking for the fake star path here?  I have not run fake stars yet.  I need to make the model grid first. 

I got around this for now.  I put a pointer to a dummy file (but one that exists) in for the astfile parameter, and it seems to be running now.  May want to fix this so that if it does not need asts to do what is being asked to do (e.g. make the model grid), it will not require that file to exist.

karllark:
This should be fixed when we split up the run_beast.py code to run each component separately for v2.0.  Will leave this open to make sure this happens.

karllark:
What might be better is to move the ASTs input list generation from the run_beast.py code to a separate script.  Might happen naturally as we update the way to run beast scripts.

karllark:
Check out the closed (but not merge) pull request #196 for code on doing this.

PR #196, opened 2018-04-27
Beast production example

cemurray3:
Renamed example folder and edited pathnames in datamodel.py and run_production_model.py from LMC/METAL to PHAT. Edited ./beast/tools/verify_params.py to allow for physics model runs without requiring AST files.


cemurray3:
This PR addresses issue #95

karllark:
So, is this pull request ready?  Or is there more work to do before merging?  I've added the WIP label for now as I'm not sure.

lea-hagen:
I've been working on the METAL production runs and have a suggestion.  I have a bunch of fields (35+), so with `run_beast_production.py` in its current form, I would have to manually change `basename` every time I run it.  I've written my own version to take `basename` as an input, which also means I need to change the jobs written by `setup_beast_batch_fit.py`.  I'm wondering if these changes would be useful to incorporate here, or whether my use is too specialized.

karllark:
Sounds like a good idea.  If nothing else, you could create a new example subdir and have the super-ninja example for METAL.  :-)

karllark:
@lea-hagen or @cmurray-astro : any plans to create a new pull request to address the issues that this pull request would have closed?  Just checking...

lea-hagen:
I don't have any immediate plans.

Issue #121, opened 2017-09-25
Sparse likelihoods may have repeat values


karllark:
At an ISM meeting, how the sparse likelihoods are being sampled was discussed.  The current method could have repeat values.  Solution is:
remove weights
use replace=False keyword.

In addition, if fewer than size of points, deal with this either remove replace=False or do something else.

PR #200, opened 2018-04-27
updating the sparse likelihood sampling

lea-hagen:
Made changes following #121:
* removed weights
* added `replace=False` (so there's sampling without replacement)
* if the number of samples is larger than the number of points, take the likelihood as-is


karllark:
Close/open pull request to trigger a *new* travis build using the latest master.
Restarting the travis build on the travis site, just restarts the already run build using the master as it existed for the original build.

PR #201, opened 2018-04-27
PIP installable version (try #3)

karllark:
Update the beast so it can be installed via pip. Locations for library files other than in the src directory now possible (see docs).
Useful in general and needed for megabeast work (testing, docs, etc.).

Addresses issue #155.

Using atom editor, so some PEP8 changes for extra characters automatically done.


karllark:
Not needed.

PR #206, opened 2018-05-01
batch script setup codes: use command line or functions

lea-hagen:
The `setup_batch_beast_trim.py` and `setup_batch_beast_fit.py` codes were both formatted to be called from the command line.  I wanted to be able to call them from within other python code, so I rearranged them to also be callable as functions.


karllark:
Great idea.  Is this pull request ready to merge?  Or are you still working on it?  Just checking.

Issue #189, opened 2018-04-04
Beast runs over subgrids + merge tool


drvdputt:
In light of parallelizing the computation, both for speed as well as memory reasons, it would make sense to do multiple runs in parallel for a set of sub-grids and combine the results at the end. Something similar has already been suggested in #38.  

_Here is an elaborate example for a part of the hypothetical merge:_

To merge some of the statistics, the expectation values for example, I think 1 extra value needs to be stored in the stats file (for each star, to be noted). This value is `sum(exp(lnp_i))`, which is the total, non-normalized weight of the grid, with `lnp_i` the log probabilities. Of course, the `lnp_i` values are typically too large to be able to calculate this value directly, but we can apply the following trick:

In the current code, the weights are first rescaled with an exponent `N_i`:
  `w_i = exp(lnp_i - N_i)`  

After that the sum of the weights `s_i = sum(w_i)` can be taken, and the `w_i` are normalized to 1.
  `w_i /= s_i`  

Therefore, `N_i + s_i = log(sum(exp(lnp_i))) = n_i`, which is the value we're looking for.  The partial expectation values `E_1` and `E_2` for some quantity `E` can then be combined using
  `E_total = (E_1 * exp(n_1 - N) + E_2 * exp(n_2 - N)) / (exp(n_1 - N) + exp(n_2 - N))`  

Where the exponents have again been rescaled with a common term `N`, to make the calculation tractable.

So to merge the results of fits on partial grids, we just need to store this value for each subgrid.

I'm not quite sure what happens when priors are involved though. I can see problems happening when the prior weights depend on the bin widths of one of the parameters.

drvdputt:
I have been doing some more development on this (https://github.com/drvdputt/beast/tree/subgridding) and wrote some code that splits the spectral grid into pieces after the stellar priors have been applied. Then, each of these spectral grid pieces can be turned into extinguished SED grids individually (in a custom run script).

This leaves me with a bunch of sub-sedgrids, for which the noisemodel, trimming and fitting can be done individually, and results in separate lnp, stats, and pdf1d files. I'm now looking for a way to merge the results.

A basic workflow is described in this post: https://github.com/BEAST-Fitting/beast/issues/38#issue-209310882. There is one issue which is not addressed there however: the bin values depend on the minimum and maximum value of the relevant quantity. These can be different for each subgrid (especially the those for the fluxes), and therefore we can not simply sum these 1dpdfs.

Some solutions I have in mind:
[...]

PR #209, opened 2018-05-09
tools for subgridding

drvdputt:
#189 

The current production example allows the user to run the beast in parallel for subsets of stars, but this means that the whole grid needs to be loaded for each process.

This pull request introduces a number of tools and changes that allow per-subgrid parallel processing:
- Function for splitting the grid into chunks of equal size.
- Modifications to fit.py to make the output of the subgrid runs compatible.
- Function that merges the individual pdf1d and stats files

This way, beast processes can be run on the same machine, without the memory overhead.

Most of the tools are in ``beast.tools.subgridding_tools``. Maybe there is a better location, as most of the current tools are actually scripts, instead of utility functions.

A run script is not yet included, but I have one available is that is of interest. Tips on how to modify the example run script are given in ``docs/subgrid_parallelism.rst``.


karllark:
Great to see you wrote documentation as well.

karllark:
Were you planning on including an example model run for the subgridding?

drvdputt:
I think I can provide a new output file. Although now I'm wondering why adding the distance didn't give an error for the testing, since the reference file wouldn't have had those columns.

As for the example, the phat_small data and datamodel can be used, but with a new run script. I think adding this script will be useful to show how the parallel processing works. I currently have such a script for my own use, which also contains a custom --ast section.

I think I'll add it as-is, and then we can polish it up later.

karllark:
sounds like a good plan

drvdputt:
The tests succeed when I ignore the new column by changing
`for tcolname in table_new.colnames:`
to
`for tcolname in table_cache.colnames:`
in `test_fit_grid.py`. Which is good news.

@karllark, I'll send you then new file by email.

drvdputt:
I think the current state is good enough for a merge. 

I will be using this new functionality in the nearby future, so then I'll see how it does in production.

karllark:
Ok.  Going to merge.  The failing test should go away when merged into the master (I think).

Issue #207, opened 2018-05-02
trim batch setup can make too many job files


lea-hagen:
If `num_subtrim` in `setup_batch_beast_trim.py` is set to a number larger than the total number of catalog subfiles, it will create empty job files.  Unless there are objections, I'll just force `num_subtrim` to be less than or equal to the number of subfiles.

There may be a similar issue for `setup_batch_beast_fit.py`, but I haven't checked yet.  I'll update that too if needed.

karllark:
Sounds like a good plan.

PR #211, opened 2018-05-10
minor fixes for setup_batch_beast_trim.py

lea-hagen:
Fixes two problems:
* making empty job files when `num_subtrim` is larger than the number of catalog files (#207)
* the resulting script breaks because of issues with relative imports


lea-hagen:
This is ready to be reviewed/merged.

PR #214, opened 2018-05-25
AST input list generation taking source density and background density maps into account

benw1:
Not quite perfect yet.  This generates usable lists, but needs to be tested as far as number of model copies being included in each region of interest.


karllark:
I strongly endorse merging the two functions.  Less code is *much* easier to maintain.

beast/examples/phat_small/run_beast.py
```diff
@@ -138,19 +140,22 @@
 
             # max. mags from the gst observation cat.
             mag_cuts = min_mags + tmp_cuts
-
+        print(modelsedgrid)
```
karllark: Is this needed?

drvdputt:
While trying to run some ASTs, I noticed that some of them were missing in the output of dolphot. I suspect it's because dolphot ignores stars with positions that fall outside of the image.

I once made a patch preventing negative pixel coordinates during the AST position sampling. We need something similar which takes the maximum pixel coordinates into account.

Something like **https://github.com/BEAST-Fitting/beast/commit/b95f50af93a90a287147682197fadbe73cc0465e**

lea-hagen:
@karllark Can this be merged?  @benw1 and I determined that we need this version to proceed with the PHAT analysis.  I vaguely recall that the outstanding issues for this pull request can wait to be fixed until later.

karllark:
Yes.  As you say, the minor issues can be addressed later.  Hopefully you can test out the code as part of your work and confirm all is good.  And make any necessary changes of course.  :-)

PR #215, opened 2018-05-29
option for `nice` level in trimming/fitting

lea-hagen:
I need to use a particular `nice` level for running code, so in case others need to do similarly, I added options in `setup_batch_beast_trim.py` and `setup_batch_beast_fit.py` to make the python commands in the job scripts have a given `nice`.


karllark:
Is this ready to merge?

lea-hagen:
Yep

Issue #179, opened 2018-02-24
Add systemic velocity of galaxy


karllark:
The systemic velocity of a galaxy can affect narrow band photometry (or in rare cases, wider band photometry).  Adding in an optional parameter to input the galaxy velocity would allow the physicsmodel calculation to shift the spectra appropriately before integration over photometric bands.

This really only is needed for the stellar spectra.  For completeness, it could also be done for the dust extinction, but not clear if this is really needed.

PR #217, opened 2018-06-01
Add redshift as an optional parameter

karllark:
All commits needed to merge Meredith's add_reshift into the updated BEAST


karllark:
Closes issue #179.

Issue #120, opened 2017-09-25
Handle 1D PDFs with negative values


karllark:
With the use of ASTs reporting fluxes (and not just magnitudes), the 1D PDF code defaults to a 1D pdf filled with zeros (in the workflow branch - solution instead of erroring out).  The issue is with 1D PDFs that need to be created with a log spacing for computational reasons.  This only happens for the case of the model predicted fluxes when they become negative due to the AST bias term (which is reality).

This was discussed in ISM group meeting and various possible solutions suggested including:
luptitudes (luptun & gunn 1999)
or
symlog numpy
or
asinh 

Additional info: http://www.sdss3.org/dr10/algorithms/magnitudes.php

lea-hagen:
After briefly looking into this, here are a few notes:
* The idea for this issue is to make the PDFs for fluxes, and fluxes can be negative, so a simple log isn't optimal.
* It is unclear where in the code the log is being taken.  It may be in beast/fitting/fit.py (lines ~300-375), or perhaps in beast/fitting/pdf1d.py (lines 38-41).
* next thing to do: figure out where the log happens, and make sure the new scaling only affects fluxes, and not masses (which are also log scaled).

PR #218, opened 2018-06-01
dealing with PDFs for negative fluxes

lea-hagen:
Addresses #120:
* for the fluxes that might be negative, does the symmetric log, rather than log10 (which meant removing negative fluxes from consideration - bad!)
* the `ignorebelow` option in `pdf1d.py` only existed to ignore the negative fluxes, so it has been removed (as well as references to it in `fit.py`)

A paper about the symmetric log is [here](https://pdfs.semanticscholar.org/70d5/3d9f448e6f2c10bd87a4a058be64f5af7dbc.pdf).  The formula is
`y = sign(x) * log10(1 + |x/C|)`
where `C=1/log(10)` ensures a smooth function.

Here's a plot: orange is a standard log10(x), which drops to -infinity at x=0, and blue is the symlog, which deals nicely with x <= 0.
screenshot

### DIFF
--- a/beast/fitting/fit.py
+++ b/beast/fitting/fit.py
@@ -201,7 +201,7 @@ def Q_all_memory(prev_result, obs, sedgrid, ast, qnames_in, p=[16., 50., 84.],
     ast: beast noisemodel instance
         noise model data
 
-    qnames: list of quantities or expresions
+    qnames: list of quantities or expressions
 
     p: array-like
         list of percentile values
--- a/beast/fitting/fit.py
+++ b/beast/fitting/fit.py
@@ -316,16 +316,10 @@ def Q_all_memory(prev_result, obs, sedgrid, ast, qnames_in, p=[16., 50., 84.],
         qnames.append('log'+cfilter+'_wd_bias')
 
     # create the full model fluxes for later use
-    #   save on log format like the other fluxes
+    #   save as symmetric log, since the fluxes can be negative
     full_model_flux = _seds + ast_bias
     logtempseds = np.array(full_model_flux)
-    indxs = np.where(full_model_flux > 0)
-    if len(indxs) > 0:
-        logtempseds[indxs] = np.log10(full_model_flux[indxs])
-    indxs = np.where(full_model_flux <= 0)
-    if len(indxs) > 0:
-        logtempseds[indxs] = -100.
-    full_model_flux = logtempseds
+    full_model_flux = np.sign(logtempseds) * np.log10(1 + np.abs(logtempseds * math.log(10)))
 
     # setup the arrays to temp store the results
     n_qnames = len(qnames)
--- a/beast/fitting/fit.py
+++ b/beast/fitting/fit.py
@@ -369,12 +369,6 @@ def Q_all_memory(prev_result, obs, sedgrid, ast, qnames_in, p=[16., 50., 84.],
 
         # setup the fast 1d pdf
 
-        # need to know so 'zeros' (defined at -100) are ignored
-        if '_bias' in qname:
-            ignorebelow = -99.99
-        else:
-            ignorebelow = None
-
         # needed for mass parameters as they are stored as linear values
         # computationally, less bins needed if 1D PDFs done as log spacing
         if qname in set(['M_ini', 'M_act','radius']):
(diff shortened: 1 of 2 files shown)


lea-hagen:
Here are some plots I made to demonstrate the difference in the 1D PDFs before/after making this change.  (As a side note, trying to format them nicely with matplotlib proved to be annoyingly difficult, so please excuse the clunkiness.)

I used data from one of the METAL fields.  From the catalog, I took a selection of stars with three measurements below 0 (`F225W_RATE`, `F275W_RATE`, `F336W_RATE`) and three measurements above 0 (`F475W_RATE`, `F814W_RATE`, `F160W_RATE`).  This gave me a fairly long list, so I modeled 5 of them.  I followed the identical production run procedure for both, with the exception of the `log10(flux)` vs `symlog(flux)` for the 1D PDF.

In the plots below, each row corresponds to one of the stars, and each column corresponds to the tag `log[filter_name]_wd_bias` (referring to the model SED with dust+bias).  The values of the y-axis run from 0 to 1: the PDF is plotted as PDF/max(PDF).  The x-axes between the two sets are different because the original one is simply `log10(flux)` and the new one is `symlog(flux)`.

Here are the three main observations I make between the two plots:
- Using `symlog`, there exists 1D PDFs for filters that didn't have them before!
- For the filters that have PDFs in both, the shapes and relative values are broadly similar
- You would expect that any positive values in the symlog PDFs would also appear in the log10 PDFs, and that stars with no PDF in the log10 method would only have negative values in the symlog PDF.  This appears to be the case.  For example, in F336W, stars 2 & 4 are the only ones with positive components in their PDFs for the symlog method, and they're also the only ones with PDFs in the log10 method.

I conclude that this procedure is reasonable.

### Original code (`log10`)
[compare_pdf_orig.pdf](https://github.com/BEAST-Fitting/beast/files/2065602/compare_pdf_orig.pdf)

### New code (`symlog`)
[compare_pdf_new.pdf](https://github.com/BEAST-Fitting/beast/files/2065603/compare_pdf_new.pdf)

karllark:
Thanks for the plots.  They do look better.

karllark:
The tests are not passing.  Not passing tests are expected as the bias flux columns (logxx_wd_bias) are expected to be different.  But there are other parameters that are different as well and we need to understand this.

For example, the best fit values in the stats file have changed.  This is a big change and we need to understand.  The difference seems to  be that the stats values now in linear space and before they were in log space.  
E           AssertionError: 
E           Arrays are not equal
E           logHST_WFC3_F275W_wd_bias_Best columns not equal
E           (mismatch 100.0%)
E            x: Column([7.683297e-20, 7.642224e-20, 7.505806e-20, 7.201558e-20,
E                   6.948957e-20, 7.428622e-20, 1.998531e-19, 7.476472e-20,
E                   1.158804e-19, 6.967845e-20, 7.051954e-20, 1.158804e-19,...
E            y: Column([-19.114452, -19.11678 , -19.124603, -19.142574, -19.15808 ,
E                   -19.129092, -18.699289, -19.126303, -18.93599 , -19.156902,
E                   -19.151691, -18.93599 , -19.155653, -19.128096, -19.152238,...


And there are small changes in the logflux_nd values.   This change may be the usual frustrating small change that we have not been able to track down.  But it might be something else as well.
E                       AssertionError: 
E                       Arrays are not equal
E                       testing grid/logHST_ACS_WFC_F475W_wd
E                       (mismatch 0.39987980491413566%)
E                        x: array([-17.707954, -17.707235, -17.703516, ..., -21.969072, -21.774811,
E                              -34.945572])
E                        y: array([-17.707954, -17.707235, -17.703516, ..., -21.969072, -21.774811,
E                              -34.945572])

lea-hagen:
I was thinking about the stats file, too, and I'm not sure about the best way to proceed.  If flux values are negative, and especially if the best values are negative, we can't have the output be log units.  Do we rename the variable?  Make everything in the stats file linear?  Something else?

For the logflux_nd values, it seems in the regime of the annoying small change, but obviously it's hard to say for sure.  I guess it's all on the linux machine, rather than being the linux/mac discrepancy, right?

lea-hagen:
Per discussion yesterday, I changed the keywords for the 1D PDF and stats files from `log[filter]_wd_bias` to `symlog[filter]_wd_bias`.  I also opened an issue (#222) for discussion about whether we want such a large assortment of units for the output files.

karllark:
Great - going to be merge and then update the cached files for the travis testing.

Issue #226, opened 2018-06-22
symlog update breaks merging


drvdputt:
The bins for the 1dpdfs were no longer the same (for the same reason as my comments in #189)

To gather the right parameters for the pdf1d ranges, I had to manually put in the log total fluxes.

Since these have been replaced by the symlog described in #218, a fix is necessary.

For the long run it would be better to avoid duplicating this code, but for now I have a patch which provides the correct symlog bounds.

PR #227, opened 2018-06-22
Change log to symlog in merging function

drvdputt:
Some duplicate code in the subgrid_info function already came back to bite me.

Here's a quick fix for #226.


karllark:
@lea-hagen Look good?

Issue #225, opened 2018-06-22
Bright-end mag_cuts for ASTs


drvdputt:
@benw1 and I noticed that many of the ASTs picked by pick_models and similar functions can be way brighter than what is in the catalog, and they are therefore useless.

The problem does not appear on the faint-end, since the mag_cuts parameter is used to cut out the faintest models.

We need a similar parameter for the bright end of the model grid. 

Note that due to the distances, the model grid can contain bright stars that are very close. If there are no such stars in the catalog, it does not make sense to perform ASTs for that model range.

PR #228, opened 2018-06-22
Brightness cutoff for AST sampling

drvdputt:
See #225 

This should do the trick. Models will only be used for asts if at least N of their fluxes (same N as for the faint cut) are above the given values.

Not yet used in the run files, but should be a straightforward modification.


karllark:
Makes sense once I got my brain around the fact that the checks are done in Vega mags.

karllark:
@benw1 Look good to you?

karllark:
looks fine to me, so I'm going to merge.

Issue #224, opened 2018-06-20
Example fit plot changed


karllark:
The example fit plot from the phat_small run has changed.
The values have changed slightly.
The stellar+dust+bias line is missing (likely related to the new symlog used for the pPDF 1d for these fluxes).
Compare to http://beast.readthedocs.io/en/latest/install.html#running-example

![beast_example_phat_ifit_starnum_0](https://user-images.githubusercontent.com/1554536/41679589-b54725b0-749c-11e8-950b-3b026957e1b0.png)

PR #229, opened 2018-06-26
Adapt plot_indiv_fit to symlog update

drvdputt:
I've taken a look at #224, and the plots are indeed being skipped due to the name change. That part should be easily fixed.

I'm not quite sure what the correct values to plot would be however. For the 'log' values, a power of 10 is taken before plotting them. So I tried to invert the symlog formula. Turns out that using 'expm1' (instead of 'exp() - 1') is pretty crucial. The blue curve seems to have reasonable values, but is not entirely the same though.

figure_1

### DIFF
--- a/beast/plotting/plot_indiv_fit.py
+++ b/beast/plotting/plot_indiv_fit.py
@@ -9,8 +9,6 @@ Plot the individual fit for a single observed star
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
-import argparse
-
 import numpy as np
 import matplotlib.pyplot as plt
 import matplotlib.gridspec as gridspec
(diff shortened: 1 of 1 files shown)


karllark:
@lea-hagen : Is inverting the symlog formula the right way to go here?  You are our symlog expert.  :-)

lea-hagen:
Yes, inverting the symlog is correct.

I'm going to work out the math here to make sure I follow the changes (and so I can be corrected if necessary!).

Inverting the symlog (eq 5 [here](https://pdfs.semanticscholar.org/70d5/3d9f448e6f2c10bd87a4a058be64f5af7dbc.pdf)):
`x = sign(y) * C * (10^|y| - 1)`
The exponential can be rewritten using the function @drvdputt mentioned
`10^|y| - 1 = exp( |y| * log(10) ) - 1 = np.expm1( |y| * log(10) )`
Putting it all back together, using `C=1/log(10)`:
`x = sign(y)/log(10) * np.expm1( |y| * log(10) )`

If I'm following the new code correctly, I think it's missing the `sign(y)` and the `|y|`.

lea-hagen:
@drvdputt does that fix the plot?

drvdputt:
Unfortunately, it doesn't. I just put in the correct formula (with abs and sign), but these didn't seem to matter in this case.

I'm not sure why it looks different though. Some of the values/numbers on the plot are also different, as mentioned in #224. There might be a common cause, but I don't really have a clue as to what.

karllark:
Some of the numbers should be different as the systemic velocity of the galaxy is now included.  This slightly shifts things so the numbers will change slightly.

karllark:
The wider "uncertainty" region may be due to better sampling of the 1D pPDF in the new symlog histograms.  Not sure, but that would be my guess.

Issue #16, opened 2016-12-21
Cleanup code


karllark:
Need to really go through the code and clean it up.  There is significant orphaned code that is no longer used due to changes in the design of the BEAST.  Less code will make it easier to maintain.

karllark:
Need to get the regression testing running in the automated python testing.  Once this is done, then the test coverage will provide a good starting point for identifying code that is not used.

karllark:
Regression testing has shown that eztables may not be used at all.  Need to confirm and then remove this code.  Reducing the amount of code to maintain would be very good.

karllark:
Once we are "happy", we should remove the 'old' directory.  
The old code will always be retrievable in the git history.

karllark:
Too generic.  Specific issues needed.

PR #232, opened 2018-07-04
Cleanup noisemodels

karllark:
Removed code that is not used (not updated and no current use case).
PEP8 changes as well.
Some documentation added to describe the three possible noise models.  More documentation needed, especially how to use/activate the different models.


lea-hagen:
Looks good, but two quick questions:

`perCameraASTs`: We don't currently have a use case, but is it possible that it would still be useful for other people?  Or is it really just the same as `MultiFilterASTs` but with separate input tables?

Documentation: were you planning to add more info about the noise models now, or sometime in a future pull request?

karllark:
We have never used perCameraASTs to my knowledge.  And it is just a way to input AST results in separate files instead of in one file to my understanding.  Hence, why I removed it.
More documentation would always be good.  Anything in particular you would like me to add?  I may try and get the API documentation (docstrings) working for at least.  A good time to request docs even if I just end up opening a new issue.

karllark:
Partially addresses issue #16.

lea-hagen:
Is there an example that includes the `Trunchen` noise model?

karllark:
Not that in the repository.  I should have one around.  I'll put in a issue for this.

karllark:
Thanks.

Issue #234, opened 2018-07-17
keeping ASTs within image boundary


lea-hagen:
As @drvdputt noted in #214, there is code to prevent ASTs from having negative x/y coordinates, but nothing to stop the coordinates from being too large.  My initial idea was to use the header of the reference image to define the boundaries, but for the PHAT half-bricks, the image is large enough that it would be optimal to read in just the header, and not the entire fits file.  However, if the header is saved independently of the image with `astropy`, the original image dimensions are not maintained in the header.  @benw1 proposed in an email that we use the photometry catalog to set the maximum x/y, which seems like a good idea to me.

karllark:
It should be straightforward to just read in the header, no?  Maybe using the memmap  keyword?

http://docs.astropy.org/en/stable/io/fits/

lea-hagen:
Yes, reading in the header is fine.  To be more clear, the issue is that when the header is being written into its own file, the `NAXIS*` keywords are automatically generated by astropy.  So when there's no data array, those keywords don't exist.  As far as I can tell, there's no generic way to keep that information.

However, I think this is going to be a moot point.  There is already code [here](https://github.com/BEAST-Fitting/beast/blob/master/beast/observationmodel/ast/make_ast_xy_list.py#L235) that checks the x/y values.  I haven't yet tracked down why there are still ASTs generated outside that range.

lea-hagen:
I believe I have located the problem: it's an issue with `pick_positions_from_map` in `make_ast_xy_list.py`.  The x/y positions from `pick_positions` are fine, but when the source density map (or background map) is used in `pick_positions_from_map`, the RA/Dec are used to define the min/max coordinates (see [here](https://github.com/BEAST-Fitting/beast/blob/master/beast/observationmodel/ast/make_ast_xy_list.py#L112)).  Unless the image is perfectly oriented north-south, the source density maps are guaranteed to have some pixels that extend beyond the edges of the image, since the pixels are evenly spaced in RA/Dec and not in x/y.

I see two possible options:
1. Change the source density and background maps to have pixels oriented along the input x/y.  If the input catalog only has RA/Dec, this would require the input image/header for WCS info.  Or perhaps make the x/y orientation optional, and check in the AST code what the orientation is?  That's sounding pretty complicated.
2. Add the catalog as an input to `pick_positions_from_map` (it's already an input to `pick_positions`).  If the x/y info exists, use the min/max values as an additional constraint.

I'm going to opt for the second option.

[...]

PR #236, opened 2018-07-24
WIP: Checking AST coordinates

lea-hagen:
This is an update to `pick_positions_from_map` in `make_ast_xy_list.py`, following discussion in #234.  It now takes the source catalog as an input (which is also updated in `run_beast.py` for the `phat_small` example).  Initially, I just checked against the min/max values for x and y, but then changed it to use a procedure that doesn't assume a rectangular image.  I also added the HDU index as an optional input (default=1), rather than the previous hard-coded index of 1.

This is labeled as a work-in-progress because the final AST list should be processed through @benw1's pipeline to confirm that the changes are good.


lea-hagen:
Huh, didn't plan to close this.  I apparently did some of the commits for this on master, so I just rearranged everything to make my branches make sense, with the side effect of closing the pull request.  I'll put in a new pull request with the new (and clean!) branch.

PR #238, opened 2018-08-20
DensityMap class + tool to split obs/AST catalogs using map

drvdputt:
I put the functionality of binning the tiles of the background/source density maps into a single class.
- The column containing the density value is now simply called 'value' instead of having to differentiate between 'sourcedens' and 'median_bg'. This way we can do away with the 'input_column' argument of `pick_positions_from_map`.
- The binning of the tiles is done in the class implementation, which ensures that is used consistently in both `pick_positions_from_map`, and my new script, `split_catalog_using_map`.
- The density map is now stored in a hd5 format, because then I can use the `self.meta` variable of the Astropy Table to store some useful information.
- I changed the phat_small run script a bit to account for this change


karllark:
Does there need to be an update to the documentation on how to generate the ASTs?  I imagine so, but am not sure.

drvdputt:
It looks like `pick_positions_from_map` (which already existed before this pull request) has not been mentioned/discussed yet in the documentation. And it mentions `pick_models_per_background`, which doesn't exist anymore. 

So some small changes would be necessary to reflect this new name.

karllark:
@drvdputt : can you update the documentation as part of this pull request?  Much better to get code and documentation at the same time.

drvdputt:
I pushed a new commit. Aside from adjusting the documentation, I also made some small changes in run_beast. The output file could be overwritten a couple of times when using one of the map options (because there were 'if's instead of 'elif's).

karllark:
Looks good - thanks for the updates to the docs.  Will merge.

PR #242, opened 2018-08-28
Check AST coordinates

lea-hagen:
[This is nearly identical to #236, which closed because I had to untangle some branches.]

This is an update to `pick_positions_from_map` in `make_ast_xy_list.py`, following discussion in #234. 
* The source catalog is an input.  The code uses [ConvexHull](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.ConvexHull.html) to make sure that ASTs are not chosen outside of the catalog area.
* The vertices of a bounding region, `set_coord_boundary`, are an optional input to set an additional constraint on the allowable region for ASTs.  For now, they must be RA/Dec (converted to x/y with a `refimage`), though that could be updated in the future to allow x/y.  These coordinates are set in `datamodel.py`.
* The `refimage` HDU is an optional input (default=1), rather than the previous hard-coded index of 1.
* The `phat_small` example (`run_beast` and `datamodel`) has been updated with these changes.

Since I started this before Dries' changes in #238, merging will require some manual intervention.

This is labeled as a work-in-progress because the final AST list should be processed through @benw1's pipeline to confirm that the changes are good.


lea-hagen:
In testing the version that also includes the changes from #238, I found that the `DensityMap` imports in `create_source_density_map.py` and `create_background_density_map.py` needed a minor adjustment:
`from density_map import DensityMap` -> `from .density_map import DensityMap`

lea-hagen:
The ASTs for PHAT seem to be running ok, so this code is ready to be merged.

karllark:
Ok.  Merging.

Issue #240, opened 2018-08-24
condensing the spatially-reorganized 1D PDFs


lea-hagen:
The 1D PDFs that are created during the fitting have N+1 rows (N = number of stars), where the last row defines the bins for the PDFs.  The bins are based on the values of the parameters present in the individual run.  Therefore, particularly when doing a production run with files divided by flux, the range of parameters will be different between runs - most notably the fluxes and masses.  The number of bins could also be different.

When running `condense_beast_results_spatial.py`, it is assumed that the array size of the 1D PDFs is the same between the sub-files, as well as the definitions of the bins.  Since this isn't generally the case, we need to figure out a way to properly combine the information.  Also, it currently crashes if the number of bins is different.

(Fortunately the MegaBEAST uses the `lnp` arrays and not the 1D PDFs.)

karllark:
Humm....good point.  Not quite sure how to solve this.

karllark:
Would have to check the merging code for the subgrid runs, the flux pPDFs might be used to generate the stats files, so maybe they are used.  @drvdputt can you comment?

drvdputt:
Yes, these files are used in the subgrid merging code.

Let me quickly summarize what happens in `merge_pdf1d_stats()` (found in `beast.tools.subgridding_tools`):
- The bin definitions are checked (equal number and equal center values)
- A weighted sum of the PDF data is taken
- The 'Best', 'Exp', 'chi2min', 'Pmax' and 'total_log_norm' columns for the combined stats file are calculated by taking the value from the partial stats files, and and applying the right formulae.
- The percentiles for the total grid are obtained from the newly calculated 1D PDFs

So yes, this part relies pretty heavily on the specific format of the pdf1d data. 

Also, don't forget that the `plot_indiv_fit` script also uses the pPDF data, but not the flux ones. That said, I do want to mention that if the noise model and the star catalog are different for each spatial region, any of the fit parameters could have their ranges modified by the trimming process.

lea-hagen:
Just to clarify, @drvdputt, which 1D PDFs are you referring to?  There are a few different versions:
1. the files output by the fitting code
2. the files that are spatially reorganized (which are formatted similarly to 1)
3. the files that combine each of the spatially-reorganized files within a given pixel
I'm concerned about going from 2 -> 3, so I want to make sure I'm following your explanation in that context.

drvdputt:
I was only talking about the files that are output by the fitting code (1).

[...]

PR #245, opened 2018-08-30
condensing spatially-reorganized 1D PDFs: also save the bin info

lea-hagen:
This addresses #240.

In `condense_beast_results_spatial.py`, the 1D PDFs are now saved as 3-dimensional arrays: `[star_num, :, 0]` has the PDF values, and `[star_num, :, 1]` is the corresponding bin values.  The length of the second dimension of the array is set to the maximum number of bins among the stars within that pixel, and any stars with fewer bins has the extra filled with `np.nan`.

`plot_indiv_fit.py` is also edited such that in can take the normal 1D PDF files as well as these new ones.  Since the normal files have 2-dimensional PDF arrays, it just needs to check the dimensions of the array, and constructs the PDFs/bins accordingly.

### DIFF
--- a/beast/plotting/plot_indiv_fit.py
+++ b/beast/plotting/plot_indiv_fit.py
@@ -52,28 +52,34 @@ def disp_str(stats, k, keyname):
 def plot_1dpdf(ax, pdf1d_hdu, tagname, xlabel, starnum,
                stats=None, logx=False):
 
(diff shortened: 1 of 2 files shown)


karllark:
Ready for merge?

lea-hagen:
Yes

PR #250, opened 2018-09-13
Test for plotting fit results

cmurray-astro:
Added test for plots using pytest-mpl. Removed plot_indiv_fit_paper.py after editing plot_indiv_fit.py to make plots look nicer. All tests pass except for one, which has an error related to the coverage test. More investigation needed!


cmurray-astro:
Never mind, the tests seem to pass! One thing I still need to do is test whether the plots are being checked or if it's still only the code

karllark:
Check out
http://docs.astropy.org/en/stable/development/testguide.html#image-tests-with-pytest-mpl
as there does seem to be a way to do mpl tests with the astropy package setup.

karllark:
Just needs '-a "--mpl"' added to the -remote-data travis test.  I can do this if you are busy.

cmurray-astro:
It didn't work! Perhaps something to do with the required pytest-mpl plugin?

karllark:
Not sure.   Did you try the 'python setup.py ...' command on your laptop?  Did it work with the added mpl stuff?

karllark:
I ran this command using a clone of your branch.  Are there some "hidden" files `.files` that give some mpl info?  Running the file multiple times gave the same result.

`python setup.py test -a "--mpl --mpl-baseline-path=baseline" --remote-data -t beast/plotting/tests/test_plots.py`

And got the following error:
beast/plotting/tests/test_plots.py F                                              [100%]

======================================= FAILURES ========================================
____________________________________ test_indiv_plot ____________________________________
Image file not found for comparison test. (This is expected for new tests.)
Generated Image: 
	/tmp/tmpi1xihoh3/test_indiv_plot.png
--------------------------------- Captured stdout call ----------------------------------
Downloading http://www.stsci.edu/~kgordon/beast/beast_example_phat_stats.fits [Done]
Downloading http://www.stsci.edu/~kgordon/beast/beast_example_phat_pdf1d.fits [Done]
=============================== 1 failed in 3.06 seconds ================================

cmurray-astro:
Yep I get the same error. Not sure what the issue is. Previously I thought it was working because 

`py.test beast/plotting/tests/test_plots.py --mpl --remote-data `

works

karllark:
Good luck.   I tried a few things, without success.   Maybe a question on the appropriate astropy slack channel would be in order.

PR #248, opened 2018-09-13
Simulated data

karllark:
Code to simulation observations based on BEAST physicsmodel and observationmodel grids.

Also updated the trim_grid.py to fix some errors and PEP8.

Issue #251, opened 2018-09-14
Cleanup testing code


karllark:
Duplicate code is present in the various test code files.  Much easier to maintain if the duplicate code was put in a single location and imported where needed.  This is both for the file downloads and the comparision of fits, hdf5 files.

PR #252, opened 2018-09-14
Cleanup tests

karllark:
Lots of repeated code in the various tests.
Created a helpers.py module in the main test directory and put three functions in it.  Now the various tests use these functions instead of having duplicate code.  Easier to maintain and update.
This pull request needs to go in *after* pull request #248.


karllark:
Addresses issue #251.

karllark:
Merging.  Having issues with tests not passing on the master after other merges.  This update will enable easier fixing of the issue - at least I think so.

Issue #205, opened 2018-04-29
Model Atmosphere Units


marthaboyer:
What units are the model atmospheres in?  I haven't been able to find any documentation on github or in the file headers.  (I need to make sure I've got the right units for the AGB models before I add them to the library.)

karllark:
@mfouesneau, @lcjohnso: do either of you know?  I have not checked the code yet to see if we explicitly stated what the units were and thought one of you just might know.

lcjohnso:
The units are erg/s/cm^2/AA.  As reference, the Kurucz data originally stems from [STSci CRDS](http://www.stsci.edu/hst/observatory/crds/castelli_kurucz_atlas.html) for the Castelli and Kurucz 2004 data, and all library fluxes are consistent with this data source.

Secondary note: these spectra are flux (F) and not astrophysical flux (F_astro), where F = pi * F_astro.  I found this to be an issue with the Munari library data.

Issue #158, opened 2017-12-29
Docs: stellar atmosphere grids - when/how to use


karllark:
There are multiple stellar atmosphere grids that can be used in the BEAST.  The "standard" is Tlusty + Kurucz.  More are being introduced, but no documentation is given on how or when to use alternative grids.  Do they replace the "standard"?  Enhance?

meredith-durbin:
I would be happy to add some documentation on how to use non-default grids, since I just went through the process of figuring that out a few weeks ago.

karllark:
That would be great!

lcjohnso:
I'm working on this now and will assemble a PR soon.

PR #254, opened 2018-09-17
Adding Docs for Grid Inputs: isochrones, speclibs

lcjohnso:
Added documentation page for grid inputs.  Also made minor corrections to iso and stellib code and documentation.

Closes https://github.com/BEAST-Fitting/beast/issues/205, https://github.com/BEAST-Fitting/beast/issues/158, and possibly https://github.com/BEAST-Fitting/beast/issues/71.

### DIFF
--- a/beast/physicsmodel/stars/ezmist/mist.json
+++ b/beast/physicsmodel/stars/ezmist/mist.json
@@ -1,5 +1,5 @@
 {
-	"version": [1.0],
+	"version": [1.2],
 	"url": "http://waps.cfa.harvard.edu/MIST/interp_isos.html",
 	"request_url": "http://waps.cfa.harvard.edu/MIST/iso_form.php",
 	"download_url": "http://waps.cfa.harvard.edu/MIST/",
@@ -37,7 +37,7 @@
 		],
 	"Av_value": 0,
 	"defaults": {
-		"version": 1.0,
+		"version": 1.2,
 		"v_div_vcrit": "vvcrit0.0",
 		"age_scale": "log10",
 		"age_type": "standard",
--- a/beast/physicsmodel/stars/isochrone.py
+++ b/beast/physicsmodel/stars/isochrone.py
@@ -617,7 +617,7 @@ class MISTWeb(Isochrone):
         iso_table.add_column('logg', iso_table['log_g'][:])
         iso_table.add_column('stage', iso_table['phase'][:])
         iso_table.remove_columns(['log10_isochrone_age_yr', 'log_Teff', 'log_L', 'log_g',
-                                  'initial_mass', 'star_mass', 'stage'])
+                                  'initial_mass', 'star_mass', 'phase'])
 
         # Remove phot columns and unnecessary properties
         extracol1="star_mdot he_core_mass c_core_mass log_LH log_LHe log_R".split()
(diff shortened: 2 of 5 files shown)


beast/physicsmodel/stars/ezmist/mist.json
```diff
@@ -37,7 +37,7 @@
 		],
 	"Av_value": 0,
 	"defaults": {
-		"version": 1.0,
+		"version": 1.2,
```
karllark: Does this change the outputs?  There is a failure in the travis testing.  What is the difference between versions 1 and 1.2?

beast/physicsmodel/stars/isochrone.py
```diff
@@ -617,7 +617,7 @@ def _clean_cols(self, iso_table):
         iso_table.add_column('logg', iso_table['log_g'][:])
         iso_table.add_column('stage', iso_table['phase'][:])
         iso_table.remove_columns(['log10_isochrone_age_yr', 'log_Teff', 'log_L', 'log_g',
-                                  'initial_mass', 'star_mass', 'stage'])
+                                  'initial_mass', 'star_mass', 'phase'])
```
karllark: Does this change the outputs? There is a failure in the travis testing. What is the difference between 'stage' and 'phase'?

karllark:
Very cool to see these docs.
I've put in a couple of questions in the changed code?  The remote-data test on travis is failing.  Wondering if it is the "minor" changes in the isochrone stuff (not doc changes).

karllark:
Remote data fail is due to something else - like a version change in numpy or something like that.   Just merged in another pull request that successfully ran all the travis tests a few days ago.  Now they fail.  Argh!  Will have to update the cached files.  Sigh.  Would love to know why we get these ~0.5% changes - something to do with integrating to get the fluxes in photometric bands.

beast/physicsmodel/stars/ezmist/mist.json
lcjohnso: I checked this: no, this update does not change outputs.  In fact, that's why I updated the file: passing the incorrect version did nothing (no failure of webform submit), so I thought it would be best to update in case it does cause a failure in the future.  There's only one option available (1.2), so likely it is currently an unchecked dummy variable.

beast/physicsmodel/stars/isochrone.py
lcjohnso: No: this is simply a correction in the column name.  The original column name is 'phase' which is copied+renamed to stage in line 619.  The incorrect column was being deleted to clean up the data file - this is now corrected.

lcjohnso:
Bummer about the small changes, but confident it isn't me (as you already concluded).  Odd: the changes are smaller for the failed Mac test version. (?)

karllark:
@lcjohnso : it is not you.  Check out #255 where I am documenting hunting this bug.  Looks like a change in numpy versions.  And my test cleanup has helped, mainly be allowing *small* differences instead of requiring equality.  Numerical issues and this is why equality should be not be asked for I guess.

karllark:
Going to re-try these tests and see if they work better now that the master has been "updated."

karllark:
Fixed!!!!

PR #257, opened 2018-09-18
Pep8 physmod2

karllark:
PEP8-ed code in physicsmodel directory (but not subdirectories). Found some two missing imports needed (!!!).


beast/physicsmodel/creategrid.py
```diff
@@ -737,84 +744,14 @@ def calc_absflux_cov_matrices(specgrid, sedgrid, filter_names):
...
+        cov_diag[:, k] = (absflux_cov_mats[:, k, k]
+                          * np.square(sedgrid.seds[:, k]))
+        for l in range(k+1, n_filters):
+            cov_offdiag[:, m] = (absflux_cov_mats[:, k, l]
+                                 * sedgrid.seds[:, k]*sedgrid.seds[:, l])
             m += 1
 
     return (cov_diag, cov_offdiag)
-
-#=================== TESTUNITS ============================
```
lcjohnso: Are these useful/meaningful tests that should get moved over to test rather than deleted?

karllark: Not sure.  These tests just checked to see if the code could be run.  What I would call a smoke test.  The regression tests that are done comparing to cached calculations are more comprehensive.  But they do require downloading data.  Maybe having such smoke tests would be a good idea.  Thoughts?  I will not merge these changes till we finish this conversation.  :-)

karllark: One of the tests is was for old, deprecated code - so not needed.   The 2nd relied on a isochrone file that is no longer present.  It was not clear how to adapt that file without adding in the web download of the isochrone file, and this makes the test similar to the existing regression tests.  So, I don't think any of these tests are useful.  Would be nice to have simple tests that do not rely on downloading files, but I think that should be a separate issue.

karllark:
@lcjohnso : going to merge this pull request if you don't object.  Leaving in the proto-test that is marked for skipping as it does not work.  Leaving it there to provide a start for some kind of test that does not include downloading files.

lcjohnso:
Late to the party, but this seems reasonable.

PR #260, opened 2018-09-24
Subgridding tweaks

drvdputt:
- New options for the subgridding run script. See the help strings. The script has had a general rework, too.

- The `reduce_grid_info` tool in `subgridding_tools` has a new option to reduce the memory usage. The original method tries to count how many unique values a quantity has in the total grid. If the grid is very large, this can cause too much memory usage. The `cap_unique` option simply cuts of the calculation when a certain number of unique values (default 1000) has been counted. As long as the maximum number of bins for the 1d pPDFs in fit.py (default 50) is smaller than this number, this option should be safe to use.

### DIFF
--- a/beast/examples/subgridding/run_beast_subgrids.py
+++ b/beast/examples/subgridding/run_beast_subgrids.py
@@ -115,7 +115,7 @@ if __name__ == '__main__':
         parallel = args.nprocs > 1
         if (parallel):
             p = Pool(args.nprocs)
-            for r in p.imap_unordered(function, argument):
+            for r in p.imap_unordered(function, argument, chunksize=1):
                 print(r)
 
         else:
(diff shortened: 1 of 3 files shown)


karllark:
This looks great.  Would be great to have some tests for this code.  Is that something easy to add to this pull request (probably not) or something for a separate pull request (probably)?

drvdputt:
I just pushed two tests, one for `split_grid`, and one for `reduce_grid_info`.

I'm planning to write one for `merge_pdf1d_stats` too, but I need to think about how I can avoid having to do a full run for that. I can probably download the example trimmed seds and trimmed noise model, split them, and then do the sub-fits.

karllark:
Is this ready for merging?

drvdputt:
Actually, I just started writing a test for the merging function. Still not sure if this will turn out easy or hard.

I suggest that you pull these changes so @cmurray-astro, can start using the `cap_unique` option. I can then add the test for `merge_pdf1d_stats` in a separate pull request, if I can get it working properly.

karllark:
Ok.  Merging...

PR #261, opened 2018-10-05
edit plot_cmd.py name and add docs

petiay:
- Changed plot_syncmd.py to plot_cmd.py (non-synthetic CMDs can be plotted as well). 
- Updated docs/simulations.rst to include plotting details.


karllark:
FYI: The updated docs are not included in this pull request.

karllark:
Thanks!   Merging now.

PR #264, opened 2018-10-29
Encode the filter name before retrieving vega mags

drvdputt:
I just ran into this, and noticed that the fix was already there for getFlux. This is the same fix, but for getMag.

Without this, the numpy.where will fail, as the strings in the loaded file are stored in byte format, not UTF-8.


karllark:
Great.  Merging in.  Test failure not related to this - see issue #265.

Issue #269, opened 2018-11-09
Subgridding physics model needs filters?


cmurray-astro:
When testing the subgridding routines (using the example run_beast_subgrids.py in beast/examples/subgridding/) I run across the following error when generating the physics model grid (via `./run_beast_subgrids.py -p`)

`  File "/astro/dust_kg3/cmurray1/smidge/beast/tools/subgridding_tools.py", line 81, in split_grid
    sub_g.grid.header['filters'] = ' '.join(g.filters)
TypeError: can only join an iterable`

which seems to be because the grid does not have the filters included, but I am confused why the filters are needed at this stage. @drvdputt, any ideas?

drvdputt:
Ah, I see what's going on. I wrote a test for `split_grid` that splits and reassembles an SED grid, and checks if everything's intact. I noticed that it was failing because the filters were missing, so I added line 81. Of course, this won't work for a spectral grid, but the test didn't cover that.

I guess I could change it in a way that the filters are added to the subgrid header, only if a 'filters' field is already there.

For now, try commenting out that line.

cmurray-astro:
Thanks Dries! I already added a line checking for filters, so it's all good, just figured I'd check with you.

PR #271, opened 2018-11-09
Check for filters before adding them to model subgrids

cmurray-astro:
Added line to check for filters before adding them to model subgrids, referencing issue #269 

@drvdputt does this look like a reasonable change?


drvdputt:
The subgridding tests are working, so this looks fine to me.

Issue #159, opened 2018-01-08
model parameters associated with ASTs


lea-hagen:
As best I can tell, when the fake star catalog is made by pick_models in make_ast_input_list.py, there isn't anywhere that it saves the model parameters associated with each SED.  If that's the case, I'll put in a request to add that functionality.

galaxyumi:
That output format is set for the MATCH input. If the model parameter information is needed, please make a request for a separate output format.

lea-hagen:
Coming back to this while thinking about MegaBEAST verification.  We need to be able to use the ASTs to make plots showing things like completeness as a function of A_V, stellar mass, etc.

PR #272, opened 2018-11-21
save physical parameters associated with ASTs

lea-hagen:
Addresses #159 by saving the physical parameters associated with each artificial star.  The example files and documentation are also updated.

An issue that came up during this processes was the way the SED grid was being read in and processed: it was taking absurdly long to extract the grid information for a given set of indices.  I updated that part of the code to use the same procedure as the grid trimming in [trim_grid.py](https://github.com/BEAST-Fitting/beast/blob/master/beast/fitting/trim_grid.py).

### DIFF
--- a/beast/examples/metal_production/run_beast_production.py
+++ b/beast/examples/metal_production/run_beast_production.py
@@ -163,8 +163,10 @@ def run_beast_production(basename,
             mag_cuts = min_mags + tmp_cuts
 
         outfile = './' + datamodel.project + '/' + datamodel.project + '_inputAST.txt'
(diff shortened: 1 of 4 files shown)


karllark:
How much faster is the new way to extract the SEDs from the full grid?

lea-hagen:
I don't have a very quantitative answer.  When I was generating ASTs for PHAT, I let the code run overnight and it still hadn't finished, which is when I resorted to the smaller grid.  With these new changes (and a slightly different grid), I didn't time it, but it was certainly less than half an hour start-to-finish.

karllark:
Thanks.  I'll do the merge now.

PR #275, opened 2018-11-29
plot CMD color-coded by BEAST parameter

lea-hagen:
This is similar to `plot_cmd.py` (and I borrowed code from there), but instead, the points are colored based on a BEAST parameter (default is `chi2min`).

I also made a new page in the documentation that has a sentence about this, `plot_cmd.py`, and `plot_indiv_fit.py`.  Those were the only things in the plotting folder that have been edited recently, so I'm not sure if the others are being used.


karllark:
Travis errors unrelated to this PR.  Fixed in the master.  Merging as a result.

PR #276, opened 2018-11-29
small updates to background maps

lea-hagen:
This changes how files are named, so it's easier to keep track of multiple fields.  The default is to use the full name of the reference image in the file names, and it adds an optional keyword to change the naming system to another string (which could be just the filter name).  It also saves everything into the same folder as the input catalog.  

I also made some minor changes to the plots (one of the titles, and using `plt.tight_layout`), and added a bit of documentation about the background maps.


karllark:
Travis errors unrelated to this PR. Fixed in the master. Merging as a result.

PR #286, opened 2018-12-14
regression testing for ASTs

s-goldman:
The lengths of the tables are compared rather than the values as they keep changing for some reason.

(Temporary)


karllark:
The results keep changing as the specific models picked for the AST inputs are based on a random number generator.  You will need to fix the seed to make the set picked reproducible.  I had to do this for the simobs test, you can check it out for an example of how to do this.

beast/beast/observationmodel/tests/test_simobs.py

s-goldman:
I’ve added a random seed and it should work now if you replace the old cache_inputAST.txt file that you are hosting with the one attached, and accept my pull request.

Cheers,
Steve

karllark:
Can you send me this file via email?  Was pasted into the github comment and I worry about formatting.

karllark:
Since the tests passed, I assume you'll need to update the test code once the new file is in the cache.

s-goldman:
Should work now!

PR #298, opened 2019-03-01
Adjustable stellar priors

karllark:
Addresses #43.

Full adjustable stellar priors.
Removed "hack" to get the constant SFR in logage.  Now have a proper exponential model.


lcjohnso:
Also: in investigating how the priors are set, it looks like there is an issue with the mass prior IMF calculation.  For a broken power law (like Kroupa), normalizations are required to link up the segments with different slopes.  As a result, M < 0.5 Msun appear to have incorrectly low weights.

Finally: Are grid and prior weights normalized with respect to each other?  Should they be? The absolute scaling of the age_grid_weights tied to yr/bin has a very high intrinsic normalization.

karllark:
Many thanks for the comments!  Very much appreciated.  I will investigate and respond next week once I'm back in Baltimore.

karllark:
The updates for the readthedocs is to stay current - basically RTD changed their config file formats.

kmcquinn:
**Metallicity information.** 
- Metallicity can be included in bins, the same way as age. For CMD fitting, something like a range of -2.0 to 0.1 in increments of 0.1 or 0.15 dec is generally used. 

- Stellar evolution models differ in the range they cover. PARSEC, I believe goes to somewhat lower metallicity values (-2.3) than MIST (-2.0). There is also development of models to go lower than -2.3.  In addition, some of the stellar models allow for a range in alpha/iron ratios, which I think would be great to consider in the BEAST. 

- So, ideally, BEAST would be flexible in the range of metallicities fit and allow for the secondary variable of alpha/iron ratio (which also may help constrain the SFH as different alpha/iron ratios are linked to early / late SFHs). 

- Current practices in deriving the age-metallicity relation for a stellar ensemble with CMD fitting includes have no priors, which often result in un-physical (extremely variable) Z(t) solutions, or adopting a prior such as a monotonically increasing Z(t), which is also limiting. 

- Given the individual star fits from BEAST that feed into MegaBEAST, I think there is an opportunity to probe the age-metallicity relation in greater detail, and potentially return a realistic spread in metallicity for a given age bin in the SFH, OR, potentially separate out different age/metal populations in a single field.

- Adding in stellar spectroscopic measurements in the BEAST (or potentially a metallicity distribution function in MegaBEAST?) may help reduce uncertainties in the SFH and Z(t) output from MegaBEAST. The challenge would be in how to apply constraints from individual stars into constraints for the ensemble.

karllark:
@lcjohnso : In your point 1 above - good point and is an issue for both mass and age (in converse ways).  For the age weights, in the past the prior weights were  1.0, so not mulitplying by it was ok - but not anymore.  In the case of the mass weights, the implementation of the mass_prior_weights integrated the prior across the bin, hence was a combination of the grid and prior weights already.  What really should happen is that the mass_prior_weights should be the "average" prior in each mass bin and then the mass_grid_weights gives the bin width.  I'll update the mass_prior_weights to do the correct "average" calculation so that the combination of prior and grid weights is correct (and what we were getting previously).

karllark:
@lcjohnso : you are very correct about the Kroupa IMF.  Here is a plot showing the current Kroupa and Salpeter IMFs.  Blargh.  Will fix.
imf_old

karllark:
Fixed Kroupa IMF.
imf_new

karllark:
See PR #387 for the continuation of this PR.

karllark:
For the `flat_grid` option, I've changed this to `flat_log` and given `flat` an alias of `flat_linear` to provide the equivalent and context.

karllark:
For the `bins` age case, I've changed this to be `bins_histo` and `bins_interp` to give both step like priors and "smoother" priors using bins.

PR #311, opened 2019-05-02
Updating to the new rtd config

karllark:


karllark:
Going to merge as only then is it possible to see if it fixes the online readthedocs build.

Issue #294, opened 2019-02-28
Noise model: use_rate as default option / definition of completeness


drvdputt:
Currenlty, use_rate is an option in the functions that are called to generate the noise model. It makes sure that the rates from the AST output are used, instead of the vega magnitudes. This is important for modelling negative flux measurements, as these will all have magnitude 99. Without the use_rate option, all these fluxes will be set to zero, and the bias for weak sources will hence be overestimated.

The magnitudes should only be used for legacy reasons, so the default should be changed to use_rate=True. Or equivalently use_rate should become use_mag instead, with False as default.

This also brings in a discussion point about how the completeness is calculated. When the magnitudes are used, anything that was flagged with a magnitude of 99 will be assigned a 0 flux, and considered incomplete. This means that sources with negative measurements will be considered incomplete. This is not correct, as the negative fluxes are useful measurements. Only the ASTs for which all the rates are exactly zero should be considered incomplete.

There is another problem: any preprocessing done on the AST results will change the completeness. As an example, I am currently using a preprocessing step that applies the same filter to the AST results that I used for the observed catalog. One of its criteria is to remove any stars that have zero fluxes. As a consequence, the completeness will be 1 in the noise model. 

Maybe the removed ASTs should be kept in the catalog, but with some flag added, which the BEAST could then use to calculate the completeness properly.

PR #330, opened 2019-05-07
Make using the AST rate column the default

karllark:
Addresses #294.

### DIFF
--- a/beast/examples/phat_small/run_beast.py
+++ b/beast/examples/phat_small/run_beast.py
@@ -188,7 +188,8 @@ if __name__ == '__main__':
            datamodel.noisefile,
            datamodel.astfile,
            modelsedgrid,
-           absflux_a_matrix=datamodel.absflux_a_matrix)
+           absflux_a_matrix=datamodel.absflux_a_matrix,
+           use_rate=False)
 
         # in the absence of ASTs, the splinter noise model can be used
         # instead of the toothpick model above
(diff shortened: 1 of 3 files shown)


karllark:
@lea-hagen : very simple PR, so should be quick to review.  But always good to get a 2nd set of eyes on the changes.

beast/examples/phat_small/run_beast.py
```diff
@@ -188,7 +188,8 @@
            datamodel.noisefile,
            datamodel.astfile,
            modelsedgrid,
-           absflux_a_matrix=datamodel.absflux_a_matrix)
+           absflux_a_matrix=datamodel.absflux_a_matrix,
+           use_rate=False)
```
lea-hagen: Is this `False` because the PHAT photometry is only magnitudes?

beast/observationmodel/tests/test_toothpick_noisemodel.py
```diff
@@ -37,7 +37,8 @@ def test_toothpick_noisemodel():
                             asts_fname,
                             modelsedgrid,
                             absflux_a_matrix=absflux_a_matrix,
-                            vega_fname=vega_fname)
+                            vega_fname=vega_fname,
+                            use_rate=False)
```
lea-hagen: similar question for this - I'm guessing changing to `True` would mess with tests?

beast/examples/phat_small/run_beast.py
karllark: Yep.  The phat_small and regression tests are based on *old* ASTs.  So, only output magnitudes available (no rate columns).

beast/observationmodel/tests/test_toothpick_noisemodel.py
karllark: Yep.

lea-hagen:
Great!  Looks good to merge.

Issue #359, opened 2019-07-17
discrepancies in how extinction is handled by BEAST code and dust_extinction


lea-hagen:
The BEAST has several built-in extinction curves ([here](https://github.com/BEAST-Fitting/beast/blob/master/beast/physicsmodel/dust/extinction.py)).  When input wavelengths are outside of their defined ranges, they return 0.

We've recently started switching over to the `dust_extinction` package (#259, #300).  Unlike the BEAST implementation, these exit with an error when wavelengths are outside the allowed range.

The effect of this: the stellar grids extend over a wide range of wavelengths, so when we run `make_extinguished_sed_grid`, code that used to work fine is now not working.  We've been fine in the past because our filter sets don't overlap the wavelengths where extinction curves are undefined, but this is something that should be fixed.

In the short term, we want to use `F20` from `dust_extinction` for v1.3 production runs.  I see three options (though there are perhaps more):
1. Edit `Generalized_DustExt` to have an option to return 0 for wavelengths where `dust_extinction` says the curve is undefined.
2. Edit the creation of the stellar grids such that they don't extend beyond where the extinction curve is defined.
3. Implement F20 in the BEAST's extinction curve file, and allow 0 extinction where the curve is undefined, thus temporarily not using `dust_extinction` at all.

@karllark and @lcjohnso - you may have thoughts about this.

karllark:
I favor 1 or 2.  Avoids having to maintain F20 in the BEAST.

1. Could be done as all the dust_extinction models include the allowed wavelength range as part of the model.  E.g. mod = F20(), F20.x_range where x_range is in units of 1/micron (yeah would be better at microns I know).

2. Probably the cleanest and most robust solution.  Kinda worrying to have SEDs created that have silently not applied dust extinction outside of the range defined for the extinction model.

lea-hagen:
I also like 2 the best.  I haven't worked with that part of the code before, though.  If someone else can easily make that change, that would be great... otherwise I can take a look.

PR #360, opened 2019-07-17
Fix extinction/dust_extinction discrepancies

lea-hagen:
Addresses #359:
* all BEAST extinction curve prescriptions have the attribute `x_range`, and if the input wavelength is outside of that range, it throws an error (conveniently using `_test_valid_x_range` from `dust_extinction`)
* In `beast.physicsmodel.model_grid`, `make_spectral_grid` takes the extinction curve (`datamodel.extLaw`) as an optional input.  If it's set, any parts of the spectra with wavelengths outside of `x_range` are discarded.

### DIFF
--- a/beast/examples/metal_production/run_beast_production.py
+++ b/beast/examples/metal_production/run_beast_production.py
@@ -113,6 +113,7 @@ def run_beast_production(basename,
             redshift=redshift,
             distance=datamodel.distances,
             distance_unit=datamodel.distance_unit,
+            extLaw=datamodel.extLaw,
             add_spectral_properties_kwargs=extra_kwargs)
 
         # add the stellar priors as weights
--- a/beast/examples/phat_small/run_beast.py
+++ b/beast/examples/phat_small/run_beast.py
@@ -92,6 +92,7 @@ if __name__ == '__main__':
             redshift=redshift,
             distance=datamodel.distances,
             distance_unit=datamodel.distance_unit,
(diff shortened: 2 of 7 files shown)


karllark:
This PR causes the CI testing to fail with the out-of-range error for the dust extinction model.  What needs to be changed in beast/physicsmodel/tests/test_extinguished_sed_grid.py to get back to the previous solution?

see: https://travis-ci.org/BEAST-Fitting/beast/jobs/560175889

karllark:
Maybe an updated version of the spectral grid file to cache for comparison during the tests?  And maybe an update to beast/physicsmodel/stars/tests/test_spectral_grid.py instead?

lea-hagen:
I had forgotten to update the examples, so that's done now.  Since this update fundamentally changes the SED grid, I think it would be best to update the cached version.  I created a new one by running the phat_small example -- where do I put it?

karllark:
Somewhere on central storage and tell me where it is.  I'll update the cached version then.  Wish there was a better way, but not yet.

lea-hagen:
just sent you a message with the file path

karllark:
An update to beast/physicsmodel/stars/tests/test_spectral_grid.py is needed.  It is still building the old - larger grid.

lea-hagen:
Looks like it's ok now?

karllark:
Looks great.  The current failure is an unrelated documentation issue.  So, going to merge.

PR #366, opened 2019-08-20
WIP: new user interface

s-goldman:
Added main.py script as a master script for use outside of python using console scripts. Also edited the setup.py to allow the console scripts to work


s-goldman:
@eteq I'm not sure how to use your suggested method on multiple .py files so I've opted for a function, however this gives me an error:

`error: cannot have multiple subparser arguments`

karllark:
Just as a reminder, these scripts need to move the beast/beast directory to work properly and be distributed.

eteq:
@s-goldman - this is on the right track, but to work as an entry point, `main.py` needs to be in the `beast` package directory.  You can tell this because the entry_points entry says `beast.main:main` so it's expecting to find a function named `main` in a package named `beast.main`.  I also needed to remove the `run_beast` import because it wasn't working  (where is it supposed to be?) and add a "." before the beastplotlib import, but then the command line script ran as it's supposed to.


As a related aside, you might want to rename `main.py` to `console_script.py` or something, just because `main` is a bit ambiguous of a name.

(BTW, the way I tested the above was using `python setup.py develop` or `pip install -e .` from the beast source directory.  It might be different if you do `python main.py ...`, but the pip/develop version I think is the "real" one.)

main.py
```diff
@@ -0,0 +1,59 @@
...
+from beast.plotting import plot_cmd, run_beast
+
+
+def main():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('--global-setting', action='store_true',
+                        help='some global thingie')
+
+    subparsers_run = parser.add_subparsers(dest='run', help='run help')
```
eteq:
```suggestion
    subparsers = parser.add_subparsers(help='sub-command help')
```

main.py
```diff
@@ -0,0 +1,59 @@
...
+    subparsers_plot = parser.add_subparsers(dest='plot', help='plot help')
```
eteq:
```suggestion
```

main.py
```diff
@@ -0,0 +1,59 @@
...
+            for parname, arg in inspect.signature(func).parameters.items():
+                # sanitized_name = parname.replace('_', '-')
+                if arg.default == inspect.Signature.empty:
+                    subparser.add_argument(parname)
+                else:
+                    subparser.add_argument('--' + parname,
+                                           default=arg.default)
+        return subparser
+
+    add_func(func_run, subparsers_run)
```
eteq:
```suggestion
    add_func(func_run, subparsers)
```

main.py
```diff
@@ -0,0 +1,59 @@
...
+    add_func(func_plot, subparsers_plot)
```
eteq:
```suggestion
    add_func(func_plot, subparsers)
```

Issue #375, opened 2019-09-05
pip install v1.3 is out of date


lea-hagen:
I just created a new environment and did `pip install beast==1.3`, following the instructions [here](https://beast.readthedocs.io/en/latest/workflow.html#production-conda-environment).  I attempted to import code added in #364, but it didn't work.  So I navigated into the installed code (`~/.conda/envs/beast_v1.3/lib/python3.6/site-packages/beast/`) and discovered that the newest files aren't there.  Also, the `examples` directory is empty, but that might not be related.  Anybody else having the same situation for the v1.3 release?

karllark:
That is odd.  When I check the v1.x branch, those files are there and the example directory is not empty.  But when I download the tar.gz file from pypi, those files are not there.  Argh.  I'm not sure what went wrong.  I'll work on it now.

karllark:
I found the problem.  The new `beast/tools/run/` directory did not have an `__init__.py` file.  As a result, that directory is not included in the build.  This is the same reason the examples directory is empty.  Argh.  I'm going to add __init__.py files to all those directories and tag a new release.  This release will be v1.3.2 as I only figured this out after doing a v1.3.1 release.  Exciting times!

karllark:
Ok.  I just uploaded the v1.3.2 release to pypi.  Give the pip install a try and tell me how it goes.

lea-hagen:
That fixed it.  Thanks!

karllark:
Yeah.

PR #376, opened 2019-09-05
Adding (and cleaning) needed init files

karllark:
__init__.py files needed in each directory where we want files included in the build/distribution.  Also added the needed data for the examples to setup.cfg.


lea-hagen:
Since this is related to the pypi changes in #375, could you update the [docs for production runs](https://beast.readthedocs.io/en/latest/workflow.html#production-conda-environment) to say `pip install beast==1.3.2`?

PR #380, opened 2019-09-12
Remove unused code

karllark:
Went through the code base and removed unused code from many files.
For many files, also used the 'black' command to fix many PEP8 issues.

Lots of changes due to PEP8 issues.


karllark:
Yeah!  Thanks @lea-hagen for the review.

PR #284, opened 2018-12-14
WIP: Evolutionary tracks for physicsgrid

karllark:
Adding evolutionary tracks in as an option.   Joins the current only option of isochrones.
The idea is that for single star SED fitting, evolutionary tracks may (should?) allow for a better sampling of the physical parameter space with fewer points.
To be determined.
Will address issue #267 & #18.

PR #387, opened 2019-09-26
Adjustable stellar priors (attempt #2)

### LINKED ISSUE #43: Adjustable age, mass, and metallicity priors

karllark:
Adds the ability to have adjustable stellar priors.  Previously the stellar priors were hard coded.  The goal of this PR is not to have all possible priors coded, but to have the ability to add them fairly easily in later PRs.  The current set of implemented priors is to get things started.

Brief listing of options for priors (see beast_priors.rst for more info)
- Age priors: flat_linear (also flat), flat_log, bins_histo, bins_interp, exp
- IMF priors: salpeter, kroupa
- metal priors: flat

Prior docs as built temp location: http://www.stsci.edu/~kgordon/beast/docs/beast_priors.html

In addition, all the stellar and dust priors are normalized to have an average of 1.  This avoids numerical issues.  As part of this, removed the amplitude parameters (e.g., 'N') from the dust priors as they were not actually needed (even previously), and very clearly are not needed now that the priors are explicitly normalized.  

Closes #43.

Fixes two bugs there were identified in a previous PR attempt (see below).  The Kroupa IMF is now correctly normalized between segments so it no longer has jumps between segments.   The mass prior weight no longer includes the bin weight fixing the incorrectly computed prior_weight in the grid (the weight values in the grid were correct, just this component that is not used was not).

This is a manual rebase from the previous PR #298.  Some of the comments in that PR are answered in that PR (i.e., check out that PR for responses).


karllark:
@lcjohnso : One carry over question from the previous prior.  For the exponential prior, I had thought that the shifting in time would be done by just not having the physics grid have any ages before the exponential started.  But maybe I am not understanding how exponential priors work.  Should the exponential prior be in absolute time (more like a peak)?  

Or should I remove the exponential for now and we can work on this at a later date?

karllark:
This PR needs at least one and hopefully multiple reviews, especially from stellar people to make sure the adjustable priors make sense.  More can and should be added in future PRs, but hopefully this gets us started.

karllark:
@kmcquinn : I agree with your comments about additional metallicity priors - see issue #399.  

Adjusting the min/max of the IMF is a bit more tricky, as currently we are relying on the isochrones we download to define this.  There are plans to have more control over the stellar grid - see PR #284.  More IMFs forms are always welcome and should be straightfoward to add with the capabilities introduced with this PR.  The stochasticity of the IMF at high masses is something that should be handled by the MegaBEAST and the right likelhood function (that accounts for such noise).

karllark:
I plan to merge this PR later today in prep for the BEAST HackDay tomorrow.  Any last reviews/comments?

PR #391, opened 2019-09-26
Change imports to absolute

eteq:
This PR is an attempt to do the relative to absolute import conversion.  It's not done but I'm putting it up now so that if someone else wants to finish it out they can!


eteq:
Crap it looks like I might have been doing this against a stale master, hence all the conflicts... will try to fix later unless someone else jumps in.

karllark:
Needs rebasing - too many conflicts.

PR #393, opened 2019-09-26
WIP: Absolute imports AND new user interface

s-goldman:
Added console_script.py in beast, changed the setup.py file, and changed the absolute imports for the plotting files


beast/console_script.py
```diff
@@ -0,0 +1,58 @@
+import inspect
+import pdb
```
karllark: Not used.  Delete if this is correct.

beast/console_script.py
```diff
@@ -0,0 +1,58 @@
...
+import argparse
+
+import beast
```
karllark: Not used.  Delete if this is correct.

karllark:
Would be great to include documentation on how to use this new capability.  Might need to create a new section and/or rst doc file for this.

karllark:
One great script to include in this new way would be the one that gets the library files.  Currently the command is "python -m beast.tools.get_libfiles" and it would be great it was in more like (if I understand things) "beast get_libfiles".

Issue #397, opened 2019-10-01
using Shapely/GEOS for processing regions


lea-hagen:
There are a few places where we need to evaluate the overlaps of regions.  Some examples:
* finding where some subsets of filter overlaps exist in an image
* determining whether a source density pixel is within a region (e.g., for dropping ASTs on the image)
* for image edges, calculating the area for a source density (#390)

So far, we've been using matplotlib `Path` objects. The types of operations we want to do are only partially implemented by `Path`, so some of the code has been a bit clunky.  Now I'm running into cases where `Path` can't solve my problems.

The GIS community has a package called [Shapely](https://shapely.readthedocs.io/en/stable/project.html) (which is a python wrapper for [GEOS](https://git.osgeo.org/gitea/geos/geos.git)), and it does all sorts of fancy polygon manipulations.  It looks like it would be perfect for some stuff I'm working on, but since it's another package to install, I wanted to check here before adding a dependency.

karllark:
I'm fine with adding a dependency for this.  Especially if it makes what we need to do easier!

Issue #381, opened 2019-09-12
Duplicate scripts?


karllark:
Has split_calalog_using_map.py superceded subdivida_obscat_by_source_density?

drvdputt:
I took a quick look and there is a small conceptual difference.

`split_catalog_using_map` only looks at the position of each source, with respect to a predefined map. Each source is assigned a bin number based on the source density listed by the map tile it belongs to. Whats actually being binned in the source density space is the tiles, not the sources themselves.

`subdivide_obscat_by_source_density` uses the density value that was added to the individual entries of the catalog. This is a more fine grained approach (unless the source density that was added to the catalog was derived from the mapped value anyway). This probably works fine for source density, as it is easy to estimate (only the positions of the other sources are needed).

The method that I developed to measure the background flux, does not always provide a reliable value for individual sources, so I thought it would be better to use the median of the sources in the tile. I do store both values in the catalog though (`indiv_bg` and `bin_median_bg`); maybe the fine grained approach could be useful in some cases.

**Note** A bigger difference is that `split_catalog_using_map` can also be used to split the AST catalog into parts, without having to assign density values to each entry in this catalog (so, simply using position).

karllark:
Sounds like after PR #400, we should remove split_asts_by_source_density.py.  Right?

lea-hagen:
Yes.  Once we do that, this issue can be closed.

PR #400, opened 2019-10-07
updates to ASTs and source density binning

lea-hagen:
This has an assortment of updates to the way ASTs and source density binning is done.

ASTs:
* Option to limit the x/y of artificial stars to where there are certain filters.  Currently can only do shapes that are convex polygons (i.e., not shapes with holes) because it uses a convex hull.
* Optional input for the WCS origin (0 or 1, default is 1 for fits) for the input reference image
* bug fix: if a source density bin tile is outside of any user-defined region, it is excluded from the list of tiles before placing sources.
* bug fix: there was an issue with checking if proposed x/y is within a given boundary
* use `shapely` for some of the region manipulation (#397)

`create_background_density_map.py`/`density_map.py`:
* bug fix: WCS conversions/calculations
* add option for the width of the bin, rather than just the number of bins
* for creating a source density map, add optional flag filter to exclude sources that are `[filter]_FLAG=99`

`split_catalog_using_map.py`:
* option to split catalogs into sub-files (e.g., 6250 stars per file)

`setup_batch_beast_fit.py`:
* bug fix: path to `tools/run/` files
* bug fix: moving the `pf=None` initialization to the beginning of the code

things in `beast/tools/run/`:
* I was running a test for using backgrounds instead of source densities in the noise model, and realized these aren't compatible.  Therefore, everything is adjusted to use the nomenclature of `split_catalog_using_map.py` instead of `subdivide_obscat_by_source_density.py` (#381).  I think this also eliminates the need for `split_asts_by_source_density.py`.
* `make_ast_inputs.py` runs now
* `create_physicsmodel.py`: added code to run model generating as a script (useful for large grids that require a bunch of subgrids)

`cut_catalogs.py`:
* add option for multiple flag filters
* option to return the cut catalog instead of writing it out

`production_runs_2019` example: 
* updated to be consistent with all of the changes above


lea-hagen:
@karllark do you know why the lgtm/travis builds aren't working?  I can't figure out if it's something wrong with the code.

karllark:
This is great @lea-hagen !  I'll look at the tests and see I can figure out the problem.

karllark:
No idea why LGTM failed.

Travis CI failed as I've had to update the files to support the adjustable stellar priors PR (#387) where there were changes in the prior numbers due to a bug in the grid weights and changes in filenames (sometime ago - more consistent naming by you I believe).  I also updated all the priors to be normalized to 1 on average to avoid numerical issues.  So, I think this PR can move forward, but there might have to be a quick 'fix' PR after this and the adjustable stellar priors PR is merged.

karllark:
Is this PR ready for merging?  Looks good to me so will merge if so.

lea-hagen:
There's one more thing I just remembered: `shapely` needs to be added as a package dependency.  Where do I add that?

karllark:
Is this PR needed for the 1.3 branch?

lea-hagen:
The last commit updates `setup.cfg` and `.travis.yml` with the package dependency (which addresses #397).  So this is ready to be merged!

karllark:
Of course, if 1.3 it will also go into 2.0.

PR #401, opened 2019-10-09
source density related updates (sorting by flux, removing unneeded code) and updates for slurm

lea-hagen:
Two things:
* `split_catalog_using_map.py`: sort sub-catalogs by flux before splitting (just like `subdivide_obscat_by_source_density.py` has always done)
* removed `split_asts_by_source_density.py` (#381), since it is superseded by `split_catalog_using_map`


lea-hagen:
Since this is still open, I added a couple more unrelated things:
* make the job lists from `setup_batch_beast_trim.py` and `setup_batch_beast_fit.py` executables (necessary for slurm to use them)
* added `write_sbatch_file.py` in `tools`, which will write out an sbatch submission file

@karllark I don't anticipate more random extra things in this PR, so it can be merged at your convenience.

karllark:
Great.  Will merge.

PR #407, opened 2019-10-17
WIP: Absolute imports AND new user interface

### LINKED ISSUE #336: The new user interface
There needs to be a rework of the beast user interface. We can use an entry point but it needs to point to a new python script that has arguments to run other scripts.

To run the beast we should be able to initiate commands outside of python using:
    `beast `
    `beast production`
    `beast plot_figs`
     etc.

Adding this new script to the entry point requires changing entry_points in setup.py, i.e.:

`entry_points = {'console_scripts': ['beast = beast.new_script.main:']}`

s-goldman:
Console scripts working for three plotting scripts: plot_indiv_fit, plot_cmd, and plot_filters.

Closes #336.

### DIFF
--- a/beast/plotting/make_ds9_region_file.py
+++ b/beast/plotting/make_ds9_region_file.py
@@ -2,6 +2,7 @@ from astropy.table import Table
 from astropy.coordinates import Angle
 from astropy import units as u
 
+
 def region_file_fits(input_file, col_color=None, col_thresh=None):
     """
     Make a ds9 region file out of the catalog or AST fits file.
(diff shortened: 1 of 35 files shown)


s-goldman:
Karl said: Would be great to include documentation on how to use this new capability. Might need to create a new section and/or rst doc file for this.

s-goldman:
Karl said: One great script to include in this new way would be the one that gets the library files. Currently the command is "python -m beast.tools.get_libfiles" and it would be great it was in more like (if I understand things) "beast get_libfiles".

s-goldman:
@karllark Should work now?

Issue #388, opened 2019-09-26
Ensure all imports to absolute imports


karllark:
Best practice and better for long-term code maintenance (from @eteq).

cmurray-astro:
I am working on this for BEAST hack day! (working through the code by hand)

PR #410, opened 2019-10-17
Less slick update of relative to absolute imports

cmurray-astro:
I went through and updated the relative beast imports to absolute imports. I did not notice that @eteq was already working on this, and I ran into many unforced errors and did a lot of brute-force fixing... but as far as I can tell I think it's working!


eteq:
Note: we should close #391 once this is in.

beast/observationmodel/noisemodel/toothpick.py
```diff
@@ -19,10 +19,10 @@
 
 from tqdm import tqdm
 
-from .noisemodel import NoiseModel
-from ..vega import Vega
+from beast.observationmodel.noisemodel.noisemodel import NoiseModel
```
eteq: This one is painful to look at (it's right, but I wonder if it's a sign that the `from noisemodel import *` trick should be used in `beast/observanmodel/noisemodel/__init__.py`? (cc @karllark)

(not necessarily for this PR, but as a broader task)

cmurray-astro:
Thanks, Erik! I'll fix the first one right away, will let @karllark weigh in on the second.

beast/observationmodel/noisemodel/toothpick.py
karllark: Using __init__.py is make this prettier is something to think about.  I agree, not for this PR.

karllark:
Closes #388.

PR #419, opened 2019-10-18
Evolutionary tracks for physicsgrid

### LINKED ISSUE #18: Add evolutionary tracks as a 2nd option for generating the stellar grid
The stellar grid is currently based on isochrones.  This the grid has uniform spacing in age, but non-uniform in mass.  Isochrones are best for dealing with ensembles of stars, but not necessarily for individual stars.  The issue for the BEAST is that the age spacing must be fine enough to sample the evolution of high mass stars and this severely over samples the evolution of low mass stars.  

One solution is to change from using isochrones to evolutionary tracks.  Stellar interior models create the evolutionary tracks which are then interpolated to isochrones.  The tracks are given for a set of stars with a range of initial masses.  Thus the stellar grid created from them will have a uniform spacing in mass, and a non-uniform spacing in age.  This should create a significantly smaller grid with the same effective coverage of stellar parameters.  Smaller grids mean faster run times *or* the ability to create grids with finer sampling of stellar parameters.

Basing the stellar grid on evolutionary tracks is definitely possible as this was what the original version of the BEAST did (in IDL).

Leo Girardi stated in an email to me that he has code for doing efficient/good interpolation on the stellar evolutionary tracks.  Hopefully, this is in python.

### LINKED ISSUE #267: Metrics for physicsmodel grid quality
The quality of our sampling of parameter space is not well defined.  We visually decided that the PHAT production  parameters were "good enough."  The production grid was based on isochrones as the base, hence are  non-uniform in mass, but uniform in age.  One idea to make better sampled grids is to go directly from the evolutionary tracks resulting in non-uniform age and uniform mass. 

How to decide which sampling is best?

One idea that was discussed at the 9 Nov 2018 HackDay (discussions between @karllark and @lcjohnso) would be to use the density of points on a HR diagram (Teff versus Luminosity) as a criteria, going for uniformity.  This would need to be checked in observed space (e.g., CMDs) as well given things may change fast in one observed band, but not  others.  Quantifying the density could be done by tabulating the min/max/median change in parameters/observables in for different grid creation methods.

karllark:
Adding evolutionary tracks in as an option. Joins the current option of isochrones.

The idea is that for single star SED fitting, evolutionary tracks may allow for a better sampling of the physical parameter space with fewer points.  Basically, evolutionary tracks enable variable age sampling for each initial mass.   The sampling of initial mass is user defined.  Isochrones that have variable mass spacing for each age and the mass spacing is not user settable - it is set by the isochrone website.

The interpolation of the evolutionary tracks in mass and metallicity is done based on the EEPs = equivalent evolutionary points.  There is an option to condense the evolutionary tracks for each initial mass based on user settable deltas in logL and logT.

In addition, this will enable to the BEAST to be used even with the isochrone website is not working.

Adds `plot` and `grid_metrics` as member functions of `EvolTracks` and `Isochrone`.

Only MIST evolutionary tracks are currently including in the "beast get_libfiles" command.  The PARSEC tracks should be added, probably in a separate PR.

Closes #267.
Closes #18.

Manual rebase from PR #284 (too many conflicts to auto rebase easily).


karllark:
Getting this working again. 

Recovered how to generate the MIST fits files from the ascii files downloaded from the MIST website (https://waps.cfa.harvard.edu/MIST/model_grids.html#eeps).

Update the evolutionary track plots to give more information about the regridding.  An example in the next comment.

Issue #343, opened 2019-07-16
Cannot install in fresh environment from setup.py


Stargrazer82301:
When installing beast-master via `python setup.py install`, the following exception is raised (at the end of the whole install shpeel):

```
(beastdev) replicators:beast-master cclark$ python setup.py install
The requested path 'astropy_helpers' for importing astropy_helpers does not exist, or does not contain a copy of the astropy_helpers package.
Downloading 'astropy-helpers'; run setup.py with the --offline option to force offline installation.
/Users/cclark/Desktop/beast-master/.eggs/astropy_helpers-3.2.1-py3.7.egg/astropy_helpers/git_helpers.py:120: UserWarning: No git repository present at '/Users/cclark/Desktop/beast-master'! Using default dev version.
  'default dev version.'.format(path))
/Users/cclark/Desktop/beast-master/.eggs/astropy_helpers-3.2.1-py3.7.egg/astropy_helpers/setup_helpers.py:165: AstropyDeprecationWarning: The package argument to generate_version_py has been deprecated and will be removed in future. Specify the package name in setup.cfg instead
  'the package name in setup.cfg instead', AstropyDeprecationWarning)
/Users/cclark/Desktop/beast-master/.eggs/astropy_helpers-3.2.1-py3.7.egg/astropy_helpers/setup_helpers.py:170: AstropyDeprecationWarning: The version argument to generate_version_py has been deprecated and will be removed in future. Specify the version number in setup.cfg instead
  'the version number in setup.cfg instead', AstropyDeprecationWarning)
/Users/cclark/Desktop/beast-master/.eggs/astropy_helpers-3.2.1-py3.7.egg/astropy_helpers/setup_helpers.py:176: AstropyDeprecationWarning: The release argument to generate_version_py has been deprecated and will be removed in future. We now use the presence of the "dev" string in the version to determine whether this is a release
  'determine whether this is a release', AstropyDeprecationWarning)
/Users/cclark/Desktop/beast-master/.eggs/astropy_helpers-3.2.1-py3.7.egg/astropy_helpers/setup_helpers.py:190: AstropyDeprecationWarning: Specifying the package name using the "package_name" option in setup.cfg is deprecated - use the "name" option instead.
  'option instead.', AstropyDeprecationWarning)
[...]

PR #420, opened 2019-10-18
Note in docs about numpy need

karllark:
Addresses #343.


karllark:
@Stargrazer82301 : looks good?

Stargrazer82301:
Reckon that should banish any confusion

Issue #417, opened 2019-10-18
Plotting scripts tests


karllark:
Need more tests for plotting scripts.  One exists for plot_indiv_fit (see tests subdir of plotting).  Should be straightforward to generate tests based on data already available for download (http://www.stsci.edu/~kgordon/beast/). 

Some specific scripts needing tests that should be straightforward:
- [ ] plot_cmd.py
- [ ] plot_cmd_with_fits.py
- [ ] plot_filters.py

cmurray-astro:
working on this! Connected to PR #428, still a work in progress.

karllark:
Closed by #428.

PR #428, opened 2019-10-23
WIP: Adding more plot tests

cmurray-astro:
Addressing #417 I wrote a baby test for plot_cmd.py. However, I ran into an issue with plot defaults... the two plots are different enough for the test to fail but not different enough to be relevant. An example is shown here: 

test_plot_cmd-failed-diff

You will see that I tried to set the plot settings manually, but this seems like a rabbit hole... this problem didn't come up for the only other plot test we have because plot_indiv_fit.py involves a LOT of custom plot setup. Any thoughts, @karllark?


karllark:
I am a bit surprised how different the plots are.  Is the difference shown from running the test on the same system that the cached version of the plot was made?

One other thought, I see the tight_layout() call is commented out.  Naively, I would have thought this would help.

karllark:
I wonder if we could test a different way.  What about the matplotlib based way described in https://matplotlib.org/3.1.1/devel/testing.html#writing-an-image-comparison-test?

cmurray-astro:
Yea, it is surprising :(. Originally I wrote a separate script to generate the baseline plots, and I've updated it to use previously cached plots from the test itself, but it still requires a larger-than-default tolerance to accommodate small differences in font size... which is weird! 

I tried the other matplotlib testing strategy, specifically:
```
@image_comparison(baseline_images=['plot_cmd'],  extensions=['png'])
```
but got an AttributeError based in matplotlib:
```
==================================== ERRORS ====================================
_____________ ERROR collecting beast/plotting/tests/test_plots.py ______________

    @functools.wraps(func)
    @nose.tools.with_setup(self.setup, self.teardown)
    def runner_wrapper():
>       yield from self.nose_runner()

/Users/clairemurray/anaconda3/lib/python3.6/site-packages/matplotlib/testing/decorators.py:288: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <matplotlib.testing.decorators.ImageComparisonTest object at 0x182ec0c978>

    def nose_runner(self):
        func = self.compare
>       func = _checked_on_freetype_version(self.freetype_version)(func)
E       AttributeError: 'method' object has no attribute 'xfail'
```
despite running matplotlib 3.1.1. Will keep working on this.

karllark:
Hum...no idea what the cause of that error is.

karllark:
So comparing the previously run test version of the plot with the current test version of the plot does not work?   And the same computer was used for both?  Or is it that the online travis ci version doesn't get the same answer as the cached version you created on your computer?

cmurray-astro:
I figured out the cause of the error was a lot of outdated pytest packages... which I updated and now include:

```
pip list | grep pytest
pytest                             5.2.1       
pytest-arraydiff                   0.2         
pytest-astropy                     0.3.0       
pytest-cov                         2.8.1       
pytest-doctestplus                 0.4.0       
pytest-mpl                         0.10        
pytest-openfiles                   0.3.0       
pytest-remotedata                  0.3.2       
pytest-runner                      4.2    
```

I've been running the tests on my machine, and they pass fine, but they fail in Travis (now both plotting tests are failing). Still pondering...

karllark:
Maybe try removing the text via:
https://github.com/matplotlib/pytest-mpl#removing-text

karllark:
Interesting to see what astropy does.  Seems to need reference images for the different versions of matplotlib. https://github.com/astropy/astropy/blob/d6e2a88493b743dd8d79a58766d40a47a8329e32/docs/development/testguide.rst#generating-reference-images

karllark:
Is this still a WIP?  Or is it ready to merge?

cmurray-astro:
It's ready to merge. Was planning to add more plot tests (i.e., more than just the one tiny one) if you don't mind keeping it open for a few more days, otherwise feel free to merge it and I can always add more later :)

karllark:
Happy to keep it open.  Especially if that means we get more plot tests!

cmurray-astro:
So I added more tests, including one for plot_filters.py (which I also updated to fix missing attributes and to make the plots look nicer!). As before, the tests run fine on my machine, but when I pushed them to test on Travis, it complains that it cannot find the `filters.hdf5` file:

```
beast/plotting/tests/test_plots.py:116: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
beast/plotting/plot_filters.py:38: in plot_filters
    flist = phot.load_filters(filter_names, interp=True, lamb=waves)
beast/observationmodel/phot.py:333: in load_filters
    with tables.open_file(filterLib, "r") as ftab:
/home/travis/miniconda/envs/test/lib/python3.7/site-packages/tables/file.py:315: in open_file
    return File(filename, mode, title, root_uep, filters, **kwargs)
/home/travis/miniconda/envs/test/lib/python3.7/site-packages/tables/file.py:778: in __init__
    self._g_new(filename, mode, **params)
tables/hdf5extension.pyx:374: in tables.hdf5extension.File._g_new
    ???
...
E               OSError: ``/home/travis/.beast//filters.hd5`` does not exist
```

Is this an issue you've run into before, @karllark? I

karllark:
Not this one, but something similar.  The issue is that the test environment doesn't have the library files and one of them is filters.hd5.  You will have to download it as part of the test script and tell the plotting code where it is.  Check out beast/observationmodel/tests/test_toothpick_noisemodel.py.

karllark:
Looking good.  Should I go ahead and merge?  

The decrease in coverage is due to the isochrone download not working right now.  Once that is fixed, we'll reap the coverage increase from this PR.

cmurray-astro:
Ready to merge as far as I’m concerned!

karllark:
Yeah!  Merging.

PR #430, opened 2019-10-25
simplifying/clarifying code for AST positions

lea-hagen:
`observationmodel/ast/make_ast_xy_list.py`: There were lots of nested if/then checks that were both confusing (at least to me) and excluded cases that were actually ok (such as requiring a reference image when it wasn't necessary).  So I've reorganized to calculate x/y and RA/Dec versions of each thing (position lists, convex hull boundaries, etc) separately, which makes the checks more straightforward.  I also moved the convex hull calculation to its own function, which makes the code a little cleaner.

`tools/run/make_ast_inputs`: Because of better checks in the code above, the position assignment section is now simpler.  I also added some more descriptive comments.


beast/observationmodel/ast/make_ast_xy_list.py
```diff
@@ -120,61 +121,75 @@ def pick_positions_from_map(
     # if appropriate information is given, extract the x/y positions so that
     # there are no ASTs generated outside of the catalog footprint
     colnames = catalog.data.columns
+    xy_pos = False
+    radec_pos = False
 
+    # if x/y in catalog, save them
     if "X" or "x" in colnames:
```
karllark: This probably should be: if ("X" in colnames) or ("x" in colnames).  I think the current version evaluates if ("X") or ("x" in colnames).  So should always be true as "X" is always true.  But maybe I don't understand how the *in* operator works.

karllark:
Is there an update to the docs needed?  Or is this just to make the code simpler/more pythonic?

beast/observationmodel/ast/make_ast_xy_list.py
lea-hagen: I changed this!

lea-hagen:
This code is fairly complicated, so it could be that a more pythonic way would be to split it into more functions.  But for now, it's at least more readable to me (whether that's the case for another person is a separate question).  

I updated the docs accordingly.  The code functionality isn't changed, but because there was a minor update for the necessity of a `refimage` (there are cases where it's not required), I went down a rabbit hole of editing the [AST info page](https://beast.readthedocs.io/en/latest/generating_asts.html).

The lgtm syntax error is a problem only in Python 2.7, so can be safely ignored.

If everything looks ok, this is ready to merge.

karllark:
Looks good to me.  Merging.

PR #432, opened 2019-10-29
code to plot noise models

lea-hagen:
I wrote `beast/plotting/plot_noisemodel.py` to make a similar visualization as [Figure 12 in Gordon+16](http://www.astroexplorer.org/details/apjaa2ee4f12), which plots bias and uncertainty as a function of flux.  It can do one (or more) noise models, as long as they're derived for the same SED model grid. 

Here's an example plot for one of the METAL fields.  The ubiquitous squiggles are because these noise models were made from a small set of ASTs.
noisemodel_plot

I also added a note about the code in the documentation.

### DIFF
--- a/beast/plotting/plot_noisemodel.py
+++ b/beast/plotting/plot_noisemodel.py
@@ -1,0 +1,161 @@
+import numpy as np
+import matplotlib.pyplot as plt
+import argparse
+
+from beast.physicsmodel.grid import FileSEDGrid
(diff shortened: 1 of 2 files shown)


karllark:
What is the difference between the red and black lines?  Anyway to include some kind of legend?

lea-hagen:
I was hoping to avoid adding legends, but alas, they're useful.  Now there's an optional `label` input which, if given a list of labels, will create a legend with those labels.

noisemodel_plot

PR #436, opened 2019-11-01
code to make a ds9 region file, and fix make_ast_inputs.py bug

lea-hagen:
I'd written code to make a ds9 region file out of a catalog, so I'll put it here so others can use it.  It's in `beast/plotting/`, since it's a visualization (`tools` didn't seem like quite the right place), but I'm happy to move it elsewhere.  I also made a note about it in the plotting documentation.


lea-hagen:
I sneaked in another unrelated bug fix.  `tools/run/make_ast_inputs.py` was distinguishing between `datamodel.ast_source_density_table` and `datamodel.ast_background_table`, instead of just using `datamodel.ast_density_table`.

This is ready to merge!

karllark:
Looks good.  Merging.

PR #438, opened 2019-11-06
updates to simulate_obs.py

lea-hagen:
I made two major updates to `beast.tools.simulate_obs.py`:
* move code into a function, so it can be imported OR used with argparse
* to support subgrids, the physics and noise model inputs are now lists (where each subgrid has its corresponding noise model), and stars are drawn successively from each of them.

This is fine to merge now, or wait until I've finished a workflow example of running the simulated observations through the BEAST.

### DIFF
--- a/beast/tools/simulate_obs.py
+++ b/beast/tools/simulate_obs.py
@@ -85,9 +85,28 @@ if __name__ == "__main__":
 
     # commandline parser
     parser = argparse.ArgumentParser()
-    parser.add_argument("physgrid", args='+', help="filename(s) of physics grid(s)")
-    parser.add_argument("noise_model", args='+', help="filename(s) of observation/noise grid(s)")
-    parser.add_argument("output_catalog", help="filename for simulated observations")
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


lea-hagen:
Looks like the LGTM alert is related to python 2.7, where dividing two integers would give an integer instead of a float.

karllark:
Any indication of how  to change this on the website? I looked in  the past and have not had success.

karllark:
Just checking - this still will work if you pass just a single file, right?

lea-hagen:
Using argparse, yes, it can take just one file of each.  (In confirming that assertion, I realized a couple of the `add_argument` settings needed fixing, so that's updated in the latest commit.)

Calling as a function, it currently requires a list of files.  If you think it would be useful to also take in a string, I can do a quick update.

karllark:
I think you can fix the issue  when calling with a single element instead of an array by using numpy.atlest_1d in the beginning of the function.

lea-hagen:
MAGIC!!!

Should be good to go now.

Issue #437, opened 2019-11-04
physics model grid not functioning


lea-hagen:
Something seems to be broken with the communication with the stellar tracks website.  When it creates a query, the resulting data table looks like this (i.e., not a data table):
```
# File generated by CMD 3.1 (http://stev.oapd.inaf.it/cmd) on Mon Nov  4 18:27:45 CET 2019
# isochrones PARSEC release v1.2S + COLIBRI release PR16, as in Marigo et al. 2017
# Basic reference: Marigo et al. (2017), ApJ, 835, 77
# Photometric system: <i>UBVRIJHK</i> (cf. Maiz-Apellaniz 2006 + Bessell 1990)
# O-rich circumstellar dpmod60alox40 dust from Groenewegen (2006)
# C-rich circumstellar AMCSIC15 dust from Groenewegen (2006)
# Kind of output: isochrone tables
```

I submitted the [default form](http://stev.oapd.inaf.it/cgi-bin/cmd) to check that the website was working, and it does create a table as expected.  The BEAST code was functioning normally as of Friday evening, so I'm not sure what would have changed between now and then.

@lcjohnso could you take a look?

lea-hagen:
The grid is functioning again (as of a couple days ago), so closing the issue.

PR #439, opened 2019-11-06
only use python 3 for LGTM

lea-hagen:
I decided to devote 15 minutes to seeing if I could make LGTM not give python 2 related errors, and shockingly, I think I might have found an answer.  So I've added `lgtm.yml` to (hopefully) force it to use only python 3.

Documentation:
https://help.semmle.com/lgtm-enterprise/user/help/lgtm.yml-configuration-file.html#python_setup
Example file that I copied the settings from:
https://help.semmle.com/lgtm-enterprise/user/Resources/downloads/lgtm.template.yml

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,12 @@
+# info from documentation:
+# https://help.semmle.com/lgtm-enterprise/user/help/lgtm.yml-configuration-file.html#python_setup
+# and from the example:
+# https://help.semmle.com/lgtm-enterprise/user/Resources/downloads/lgtm.template.yml
+
+extraction:
+  python:
+    python_setup:
+      # Override the version of the Python interpreter used for setup and extraction
+      # Default: Python 3 if no version is explicitly specified, and if there are no
+      # commits to the repository before January 1, 2017. Otherwise Python 2.
+      version: 3
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -3,6 +3,9 @@
 # and from the example:
 # https://help.semmle.com/lgtm-enterprise/user/Resources/downloads/lgtm.template.yml
 
+queries:
+  - exclude: cpp
+
 extraction:
   python:
     python_setup:


lea-hagen:
Nevermind, it looks like this is going to break things, and that changes should either be made on the LGTM website (presumably by @karllark?) or in the `lgtm.yml` file, but not both.

karllark:
How did this break things?   I would not be surprised that the alerts will change.

karllark:
Any indication of how to change this on the website? I looked in the past and have not had success.

lea-hagen:
Huh, apparently the box with the travis/LGTM results disappeared when I closed the PR.  From what I remember, all three of them failed (though the travis one was expected because of the grid download issue, #437).  I'll try re-opening it to see if they reappear.

I found some documentation [here](https://help.semmle.com/lgtm-enterprise/admin/help/overriding-global-project-configuration.html) about how to change configurations in the administrative tools.  It looks like the formatting might be the same as a yml file, so perhaps the settings could be moved to `lgtm.yml` for easier editing.

karllark:
Looks like the settings on the web are for paying accounts.  I can't find any way to get to the stated page in the docs.  So, we'll have to see if we can make the lgtm.yml file work.

karllark:
It does look like the lgtm.yml file worked.  At least for python.  The new errors are due to parsing for python 3 instead of python 2.  No idea why the C/C++ failed.  I've restarted it to see if it will work now.

lea-hagen:
I wonder if there are some settings for C++ that aren't the defaults but are hiding the inaccessible settings page.  I found a thing [here](https://lgtm.com/help/lgtm/api/fetch-current-config-api) about fetching those settings, but I have no idea what access token it's talking about.

karllark:
Anyway with the lgtm.yml file to just say don't check the c/c++ code?

lea-hagen:
Looking on [this page](https://lgtm.com/projects/g/BEAST-Fitting/beast/logs/rev/pr-aff94c09e037fe292e449a43d6f0aa37ccb10b2b/lang:cpp/stage:Preparing%20merge%20commit_5e5129889a7ff02945e9ded17ddcd058ebf780cd), it says the C++ build succeeds on the latest commit, but I'm not sure what the other two C++ failures are.

Issue #435, opened 2019-11-01
parameter recovery tests


lea-hagen:
I'm looking to create a version of [Fig 13 in Gordon+16](http://www.astroexplorer.org/details/apjaa2ee4f13), which shows how well physical parameters from an SED are recovered, for a given noise model.  I'm not sure if this is encompassed by the sensitivity tests (#106/#248) that already exist, or if those are specifically for simulating which filters affect parameter constraints.

karllark:
I think simulating the photometry work has been done.  But not the plotting the results and making the figure.   Would be great to have code for that.  I can point people in the direction of where the code that made the figure in the paper may be (central storage) if that would help.

lea-hagen:
Yes, that would help!  Thanks!

karllark:
I think.../astro/dust_kg/harab/beast/projects/sensitivity_tests/figs.py.

PR #441, opened 2019-11-08
many things for parameter verification

lea-hagen:
In working on parameter recovery tests (#435), I have made many updates:
* running of things:
  * `tools/run/make_trim_scripts.py`: moved the [rather unwieldy] trimming procedure over to its own file
  * `examples/production_runs_2019/beast_production_wrapper.py`: updated trimming section to use `make_trim_scripts`
  * `examples/production_runs_2019/beast_verification_wrapper.py`: an example wrapper for simulating observations, running them through the BEAST, and making plots
* plots
  *  `plot_param_err.py`: plot all parameter `p50`s against their uncertainties ((`p84`-`p16`)/2)
  * `plot_param_recovery.py`: 2D histogram comparing the simulated and recovered model parameters
  * `plot_triangle.py`: a triangle/corner plot of all the parameters (`p50`) against each other, with histograms along the diagonal
  * updated plotting docs to note these
* assorted related bugs/fixes
  * `observationmodel/observations.py`: changed all filter column names to all uppercase, because all lowercase was breaking things
  * `tools/simulate_obs.py`: wrap grid names in `str` (`np.atleast_1d` made them a `np.str`, which apparently didn't work)
  * `tools/run/create_filenames.py`: fixed a couple places that still had `_sed_` filenames instead of `_seds`
  * `tools/trim_many_via_obsdata.py`: for unknown reasons, it required the AST file name, but wasn't using it, so I removed it.  Also made sure trimmed files had the proper naming to match changes to `create_filenames` (part of #387).
  * `tools/setup_batch_beast_trim.py`: removed need for AST file name accordingly

### DIFF
--- a/beast/examples/production_runs_2019/beast_production_wrapper.py
+++ b/beast/examples/production_runs_2019/beast_production_wrapper.py
@@ -237,12 +237,6 @@ def beast_production_wrapper():
             print('\n**** go run physics model code for '+field_names[b]+'! ****')
             continue
 
-        # list of SED files
-        model_grid_files = sorted(
-            glob.glob(
-                "./{0}_beast/{0}_beast_seds.grid{1}.hd5".format(field_names[b], gs_str)
-            )
-        )
 
         # -----------------
         # 3. make ASTs
--- a/beast/examples/production_runs_2019/beast_production_wrapper.py
+++ b/beast/examples/production_runs_2019/beast_production_wrapper.py
@@ -7,6 +7,7 @@ from beast.tools.run import (
     create_physicsmodel,
     make_ast_inputs,
     create_obsmodel,
+    make_trim_scripts,
     run_fitting,
     merge_files,
     create_filenames,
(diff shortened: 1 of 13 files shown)


karllark:
Good stuff. 

I do wonder about making the filter names all uppercase and if this has impacts elsewhere.  I agree we should pick a convention and use it throughout the code and in the expected formats of the data files.

Note that making the filter names upper case means the simobs test is failing.

lea-hagen:
Hmm, not sure what to do about the uppercase vs lowercase.  If I'm remembering right, lowercase was failing for me because the photometry catalogs I get from Ben are always uppercase, so that's what's set in my datamodel file.

I fixed a couple bugs that codacy found (but strangely not found by LGTM).  Though I'm not sure if/how to fix things like `beast_verification_wrapper()` being too complex, or plotting code having dangerous default values.

lea-hagen:
Looks like the simobs test is working now, so unless there's anything else to update/change, this is ready to merge.

PR #445, opened 2019-11-15
Update how catalogs are cut

### LINKED ISSUE #434: choosing an output catalog after calculating source densities
There has been some confusion lately over what output catalog file to use after creating a source density map.  These are the two possible files:
* `[prefix]_with_sourceden_inc_zerofluxes`: photometry catalog with an additional column that has the source density
* `[prefix]_with_sourceden.fits`: same as above, but only for sources with flux != 0 in each band

Also, the background map calculations don't do any cuts with zero fluxes, and the only output catalog is `[prefix]_with_bg.fits`, which is the full catalog with two additional columns of background numbers.

lea-hagen:
This has some changes in how catalogs are cut.
* `create_background_density_map.py`: Per discussion in #434, the code that makes the source density maps no longer does anything with the sources with zero fluxes.  They were never part of the source density map calculations anyway (since they have a Vega mag of 99). The only outputs now are the SD map fits file, the SD map hd5 file, and the catalog with a new SD column.  Any cuts based on flux=0 should be done using `cut_catalogs.py`.
* `cut_catalogs.py`: 
  * After discussion with @karllark about how completeness is calculated, we realized that removing any source with flux=0 was a problem.  Specifically, a source in the AST catalog with flux=0 in all bands is an un-recovered star, which we need to keep in the catalog.  Now, if the `partial_overlap` option is selected, only sources with between 1 and `n_bands - 1` (inclusive) bands with flux=0 are removed.
  * Even sources with a negative flux are still measurements, but they're all given `[filter]_FLAG=99` by dolphot.  Therefore, if the `flagged` option is chosen, only sources with flag=99 and flux>0 are removed.
* I updated the workflow documentation to reflect these changes.

Closes #434

### DIFF
--- a/beast/tools/create_background_density_map.py
+++ b/beast/tools/create_background_density_map.py
@@ -492,11 +492,6 @@ def make_source_dens_map(cat, ra_grid, dec_grid, output_base, mag_name, mag_cut,
     # force filter magnitude name to be upper case to match column names
     mag_name = mag_name.upper()
 
-    # get the columns with fluxes
-    rate_cols = [s for s in cat.colnames if s[-4:] == "RATE"]
-    n_filters = len(rate_cols)
-
-
     N_stars = len(cat)
 
     w = make_wcs_for_map(ra_grid, dec_grid)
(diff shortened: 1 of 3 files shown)


karllark:
Looks great and nice to see the updated documentation.  Well done.  After the issue below is taken care of, I'll merge.

Can you please remove the unused local variable identified by LGTM?

PR #448, opened 2019-11-20
Fix catalog cut for ASTs

### LINKED ISSUE #447: AST catalog: problem with RATE=0 cuts
In #445, we implemented an updated method to cut the catalogs.  In particular, for ASTs, we remove all sources with RATE=0 in any band (indicating lack of full imaging coverage), but keep sources with RATE=0 in all bands (indicating a non-recovered source, which we need for calculating completeness).  However, this has the unintended side effect of retaining non-recovered sources in regions without full imaging coverage.

Is there another check we can do to ensure that we actually remove all sources without full coverage?  Or worded another way: for sources with RATE=0 in all bands, is there a column that could tell us how many images it overlaps with?

@benw1, @karllark

lea-hagen:
As described in #447, simply using the RATE=0 cuts wasn't working to properly remove sources that aren't in the full overlapping region.  I've updated `cut_catalogs.py` to now take in both the photometry and the AST catalogs.  If the `partial_overlap` option is chosen, it will do the RATE=0 cuts for the photometry catalog, find the boundary of the remaining sources, and apply that boundary to the AST catalog.  The AST catalog still has objects removed with RATE=0 in a subset of bands.  Also, `cut_catalogs` allows the AST files to be an optional input, so that users who only want to operate on the photometry catalog can do so.

Other changes:
* Moved the `convexhull_path` function (which calculates a convex hull and turns it into a Path object) from `make_ast_xy_list.py` to `cut_catalogs.py`, because otherwise there was an infinite circular loop of imports.
* Updated workflow docs
* Updated production run example

Closes #447

### DIFF
--- a/beast/examples/production_runs_2019/beast_production_wrapper.py
+++ b/beast/examples/production_runs_2019/beast_production_wrapper.py
@@ -17,10 +17,7 @@ from beast.plotting import plot_mag_hist
 from beast.tools import (
     create_background_density_map,
     split_ast_input_file,
-    subdivide_obscat_by_source_density,
     cut_catalogs,
-    split_asts_by_source_density,
-    setup_batch_beast_trim,
     setup_batch_beast_fit,
 )
 
--- a/beast/examples/production_runs_2019/beast_production_wrapper.py
+++ b/beast/examples/production_runs_2019/beast_production_wrapper.py
@@ -275,18 +275,10 @@ def beast_production_wrapper():
 
         # - photometry
         gst_file_cut = gst_file.replace(".fits", "_with_sourceden_cut.fits")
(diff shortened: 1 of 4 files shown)


lea-hagen:
@karllark If I'm interpreting the travis tests correctly, it looks like it's failing because of the isochrone download problem.  Was that intended to be addressed in #446?

karllark:
Actually, this means that the isochrone download is working again!  I set it up to fail if it worked (instead of just skipping the test) so that we would know when it was working again.  I'll do a quick PR to fully re-enable this test!

lea-hagen:
Oh, ok, that's great news!  Hooray! 🎉

PR #455, opened 2019-11-21
fixed the column name matching

### LINKED ISSUE #454: Fix column name matching in plot_mag_hist function
The current plot_mag_hist function isn't able to identify filters if the cases don't match. This could be solved by always comparing the lowered version of the column names to 'vega' so that the cases should always be the same.

christinawlindberg:
Closes #454.


christinawlindberg:
While Astropy FITS tables aren't case-sensitive, if you pull the column names from the table and compare them to a standard string, the comparison does become case sensitive.

This means that finding the column names that include the phrase "vega" becomes a case-sensitive task, but referencing the same column names later on when plotting (filter+"VEGA") isn't a case-sensitive task.

PR #456, opened 2019-11-21
NEW entry points / new main script / simple docs

### LINKED ISSUE #336: The new user interface

s-goldman:
Closes #336.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,0 +1,59 @@
+import inspect
+import pdb
+import argparse
+import sys
+
+import beast
+from beast.plotting import plot_cmd, plot_filters, plot_indiv_fit, beastplotlib
+from beast.tools import get_libfiles
+
+
+def main():
+    """Summary gets arguments for each of the scripts listed in 'scripts' 
+    and parses them for the function given in the input
+    """
+    all_funcs = []
+    scripts = [plot_cmd, plot_filters, plot_indiv_fit]
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--global-setting", action="store_true", help="some global thingie"
+    )
+
+    subparsers = parser.add_subparsers(
+        dest="subparser_name", required=True, help="sub-command help"
+    )
+
+    for item in scripts:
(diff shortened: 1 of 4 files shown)


karllark:
Please remove the unused imports identified by LGTM when you get a chance.

karllark:
Is this PR ready to merge?  If so, please remove the WIP tag in the PR title.

main.py
```diff
@@ -0,0 +1,57 @@
...
+def main():
+    """Summary gets arguments for each of the scripts listed in 'scripts' 
+    and parses them for the function given in the input
+    """
+    all_funcs = []
+    scripts = [plot_cmd, plot_filters, plot_indiv_fit]
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--global-setting", action="store_true", help="some global thingie"
```
karllark: A better help string is needed.

main.py
```diff
@@ -0,0 +1,57 @@
...
+            # make a dictionary containing the correct inputs to the function,
+            # extracted from the parsed arguments
+            funcargs = {}
+            for parname in inspect.signature(func).parameters:
+                funcargs[parname] = getattr(args, parname.replace("-", "_"))
+            func(**funcargs)
+            break  # drop out immediately, which skips the "else" below
+    else:
+        if False:
+            raise AssertionError("Invalid subparser! This should be impossible...")
```
karllark: Will never be reached as "if False" never is true.  Do you even need the "if False" line?

s-goldman:
The "unused" functions are being used, codacy is a liar. It's just that they are called as items in a list and therefore not explicitly mentioned in the text.

s-goldman:
Also not sure why the setup.py script is saying there is an entrypoint syntax error?

PR #458, opened 2019-11-21
Modify the code to reproduce the Figs. 16-18 in Gordon+2016

### LINKED ISSUE #435: parameter recovery tests

galaxyumi:
This code now reproduces the Figs. 16-18 in Gordon+2016. Useful to check whether the BEAST fits look reasonable or not. Below is an example plot using the phat_small run.

beast_example_phat_stats_param_unc

Closes #435.

### DIFF
--- a/beast/plotting/plot_param_err.py
+++ b/beast/plotting/plot_param_err.py
@@ -1,5 +1,4 @@
 import numpy as np
-import matplotlib
 import matplotlib.pyplot as plt
 from astropy.io import fits
 from matplotlib.colors import LogNorm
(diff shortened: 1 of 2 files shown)


karllark:
Minor update to docs/plotting_tools.rst needed to update the docs so that all will know the two types of plots that are made.

beast/plotting/plot_param_err.py
```diff
@@ -2,13 +2,14 @@
 import matplotlib
```
karllark: Does not look like this import is needed.  If this is true, please remove.

galaxyumi: Done!

PR #464, opened 2019-11-22
getting beast entry-points to work

karllark:
Building off the work that @s-goldman has done, I was able to get entry-points to work for the beast, but only for a couple of functions that have very simple inputs.  Other functions will have to be revised to allow for this type of calling of the scripts.  Definitely worthwhile, but not as simple as originally thought.

This PR provides info to update #456 (or supersede it depending on how we want to go).

### DIFF
--- a/beast/main.py
+++ b/beast/main.py
@@ -2,7 +2,7 @@ import inspect
 import argparse
 
 from beast.tools import get_libfiles
-from beast.plotting import plot_cmd
+from beast.plotting import plot_cmd, plot_filters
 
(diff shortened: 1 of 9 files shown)


karllark:
@cmurray-astro : have you seen this error before?  It is related to the plot tests, but I can't see how what I changed in this PR would cause this problem.

https://travis-ci.org/BEAST-Fitting/beast/jobs/615558866?utm_medium=notification&utm_source=github_status

karllark:
@cmurray-astro : ignore my question.  I figured it out right after I posted it (argh!).

karllark:
Closes #336.

PR #465, opened 2019-11-22
Completeness 1D and 2D plots

lea-hagen:
This adds code to visualize completeness.  It makes a triangle plot with completeness averaged into 2D plots for each pair of parameters.  Along the diagonal, there are plots of the 1D averaged completeness for each individual parameter (color coded by the completeness).  

Here's an example for IC1613.  There seems to be something interesting going on with mass and age... presumably some result of different masses being brighter/fainter at different stages of their evolution.  Also, unsurprisingly, R_V/f_A/Z don't impact the SED enough to show any completeness variation.

IC1613_completeness

### DIFF
--- a/beast/plotting/plot_completeness.py
+++ b/beast/plotting/plot_completeness.py
@@ -226,12 +226,12 @@ def setup_axis(compl_table, param):
 
     # mass isn't reguarly spaced, so take log and manually define bins
     if 'M_' in param:
-        col = np.log(compl_table[param])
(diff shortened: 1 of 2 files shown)


lea-hagen:
The ticks for mass and metallicity were evaluated using `log` instead of `log10`, so now they make more sense.  Unless there are any other things to change/fix, this is ready to merge.

Here's the new plot with proper axes:

IC1613_completeness

Issue #467, opened 2019-12-06
how to sample the log likelihoods


lea-hagen:
As I'm working on #373, I've been puzzling over the question about the proper parameters for sampling the log likelihoods.  Right now, we randomly select 500 from those that have have `lnP > max(lnP) - 10`.  A while back, this selection had been weighted by the probabilities themselves to ensure that we don't miss the highest probability regions, but we decided to switch to random to keep from double-counting the probabilities (issue in #121, PR in #200).

The issue I'm coming across now is whether 500 samples with a -10 threshold is enough to well-sample the probability space.  My first question is how these numbers were chosen, and my second question is how to test that they're reasonable.

lea-hagen:
I've done a test from which I conclude that there needs to be more samples and/or a smaller (less negative) threshold.

Details: I'm currently approaching #373 by adding up the probabilities within the specified range of physical parameters, and dividing that by the sum of all 500 probabilities.  For the example below, for each star, I add up all of the probs associated with M_ini>10 and Av>0.5, and divide by the total prob.  I then did an experiment to simulate selecting fewer `lnP` values, in which I randomly select half of them and do the same procedure.  Here's the plot:

![test_lnp](https://user-images.githubusercontent.com/32465297/70455401-3b36c980-1a7a-11ea-8ff7-7966282e2756.png)

There are a significant number of catastrophic failures (i.e., either the most or the least probable points are chosen, drastically changing the result), which means that we need to save more lnPs.  Also, there tend to be more points above than below the line, which I haven't figured out yet.

lea-hagen:
@karllark I was thinking of a couple additional lnP selection strategies.  I suspect you all considered them at some point, so I'm curious about why they're not being used.
* Save the 500 (or more) best fits
* Sort the lnPs and save as many as are need to accumulate X% of the total probability

karllark:
I think there are some issues here that we are not yet considering, but I don't know what they are in detail.  In discussions on this topic it has been @ktchrn and @mfouesneau that have knowledge and advice on this topic.  Maybe one of them can even point us to the relevant literature.

mfouesneau:
At that time I believe we were taking all samples above a threshold. 

You can also take the N best points but then you need to also compute and save their weights using nested sampling prescription. (I think this should be easy as we still have the density/volume of each grid point?)

karllark:
[...]

Issue #433, opened 2019-10-30
Format of stats output file


karllark:
There is a file with the BEAST PHAT fitting results that describes the contents of the stats output file.  This should be generalized and added to the docs.

https://stsci.box.com/s/oang8im5gjv1zpqvn4zonwc2qd5vps75

s-goldman:
I'm working on this

s-goldman:
PR submitted #462

Issue #373, opened 2019-08-23
Likelihood of different stellar phases


s-goldman:
It would be good to be able to get the likelihood of stellar phases (AGB stars in particular) using cuts in the outputs, i.e. selections in the  HR-Diagram space.  This could then be represented as a percentage (likelihood) in the output of the BEAST.

karllark:
This would be great.   Might be a great addition to catalog papers at a minimum (issue #333).  But also very useful for science papers.  Could also make computations for other physical parameters.  Probabilities of low and high AV stars.  Low and high Rv stars.  Young and Old stars.

karllark:
One question to resolve, is this something that would be done as part of the fitting?  Or a post processing step?  If the former, then the full nD likelihood function is available.  If the latter, then only the n samples of the likelihood could be used (e.g., 500).

s-goldman:
Results from Brick 9 from the BEAST results from PHAT. Next i'll select regions for different phases of stellar evolution. 

HR diagram                       |  Age v Mass
:-------------------------:|:-------------------------:
![](https://user-images.githubusercontent.com/32876747/65710068-37d99780-e060-11e9-8b24-b58b3499cfa0.png)  |  ![](https://user-images.githubusercontent.com/32876747/65711060-209ba980-e062-11e9-8f17-dcc85698ed45.png)

karllark:
What are these plots showing? The best fit values?  The p50 values?

s-goldman:
@karllark The best values

karllark:
You could also great density plots from the physicsmodel grid and the prior weights.  This would provide a nice visualization of the allowed physical range.  The PHAT results will give what is possible at the sensitivity of the that survey.  Making such a visualization might be really nice to have for other reasons (e.g., impact of different choices on stellar priors).

lea-hagen:
> One question to resolve, is this something that would be done as part of the fitting? Or a post processing step? If the former, then the full nD likelihood function is available. If the latter, then only the n samples of the likelihood could be used (e.g., 500).

If we calculate after the fitting, we won't have accurate probabilities for the lower-likelihood star types, because we presumably won't have any saved lnPs (or very few) that even overlap them.  Also, depending on how we select the lnPs, if we miss some high probability ones, we could have some catastrophically mis-identified star types (see plot [here](https://github.com/BEAST-Fitting/beast/issues/467#issuecomment-563330840)).

I see three disadvantages to doing this during fitting:
[...]

PR #468, opened 2019-12-09
WIP: Star type probability and related tools

lea-hagen:
This adds `tools/star_type_probability.py` to calculate the probability that a star is of a given type (#373).  Each type of star is its own function, so new functions can easily be added for more types of stars.  To address the questions in #467, there's also a function to compare the results from the full set of lnPs to a randomly selected subset of lnPs.

Getting this to work also required editing some other parts of code:
* `tools/subgridding_tools.py`: Added a `merge_lnp` function to combine lnPs across subgrids.  There's an optional `threshold` argument (like the one in `fit.py`) to eliminate lnPs much smaller than the maximum, which can help reduce file size.
* `tools/run/merge_files.py`: includes `merge_lnp` in the merging for subgrids
* `tools/read_beast_data.py`: Migrated over from `read_beast.py` in the MegaBEAST (BEAST-Fitting/megabeast#40).  It has functions to extract data from the files for the noise model, SED grid, and lnPs, and return them as dictionaries.  For the SED grid, if the parameter list is `None`, it will print out all of the parameters in the file (relevant to #433).  The code also has a function to get the parameter values associated with the lnPs.
* `tools/trim_many_via_obsdata.py`: If a trimmed file already exists, it will now skip to the next one in the input file.  (This isn't immediately related to finding star types, but it came up while I was testing a start-to-finish run with the lnP merging.)

------------------------------

The probability of being a star type is currently calculated by adding up the probabilities associated with the physical parameters (e.g., `M_ini > 10` & `Av > 0.5`) and dividing by the total probability.  This only uses the 500 sparsely sampled lnPs.  As is currently being discussed in #467, it seems like 500 may not be enough, or perhaps we need to choose the saved lnPs differently.  The issue is that the high-probability region of the grid may not be getting sampled well enough, so the star type probabilities are subject to significant fluctuations depending on which lnPs get chosen to save.  Until that's resolved, this PR will be WIP.

### DIFF
--- a/beast/tools/read_beast_data.py
+++ b/beast/tools/read_beast_data.py
@@ -99,14 +99,14 @@ def read_noise_data(
     noise_data = {}
 
     # open files for reading
-    with h5py.File(noise_filename, 'r') as noise_hdf:
+    with h5py.File(filename, 'r') as noise_hdf:
 
         # get beast physicsmodel params
         for param in tqdm(param_list, desc='reading beast data'):
             if filter_col is None:
-                noise_data[cparam] = np.array(noise_hdf[cparam])
+                noise_data[param] = np.array(noise_hdf[param])
             else:
-                noise_data[cparam] = noise_hdf[cparam][:,filter_col]
+                noise_data[param] = noise_hdf[cparam][:,filter_col]
 
     return noise_data
 
@@ -190,9 +190,9 @@ def get_lnp_grid_vals(sed_data, lnp_data):
         arrays of the SED grid parameters for the points in the lnp lists
     """
 
-    if type(sed_data) == str:
+    if isinstance(sed_data, str):
         sed_data = read_sed_data(sed_data)
-    if type(lnp_data) == str:
+    if isinstance(lnp_data, str):
         lnp_data = read_lnp_data(lnp_data)
 
     # get the keys in beast_data
@@ -206,7 +206,6 @@ def get_lnp_grid_vals(sed_data, lnp_data):
 
     # loop over the stars and extract the requested BEAST data
     for k in tqdm(range(n_stars), desc='extracting params for each lnP'):
-    #for k in range(n_stars):
         lnp_inds = lnp_data['indxs'][:, k]
         good_inds = np.isfinite(lnp_inds)
         for param in param_list:
--- a/beast/tools/read_beast_data.py
+++ b/beast/tools/read_beast_data.py
@@ -1,0 +1,142 @@
+"""
+Functions for interacting with the BEAST model
+"""
+
+# system imports
+from __future__ import (absolute_import, division, print_function)
+
+# other package imports
+import numpy as np
+import h5py
+from tqdm import tqdm
+
+
+def read_lnp_data(filename, nstars):
+    """
+    Read in the sparse lnp for all the stars in the hdf5 file
+
+    Parameters
+    ----------
+    filename: string
+       name of the file with the sparse lnp values
+
+    nstars: int
+       number of stars expected in the file
+
+    Returns
+    -------
+    lnp_data: dictonary
+       contains arrays of the lnp values and indexs to the BEAST model grid
+    """
+    lnp_hdf = h5py.File(filename, 'r')
+
+    if len(lnp_hdf.keys()) != nstars:
+        print('Error: number of stars not equal between nstars image and ' +
+              'lnp file')
+        lnp_hdf.close()
+        exit()
+    else:
+        # initialize arrays
+        # - find the length of the sparse likelihoods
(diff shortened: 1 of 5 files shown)


karllark:
I wonder if this PR should be split into two PRs.  One for the tools that merge the lnps from subgrid runs and one to address determining the probabilities of different types of stars/sightlines.  @lea-hagen : is this possible?  Having two PRs would make it easier to know which changes are for which purpose and make reviewing/accepting the PRs easier.

karllark:
I wonder if it would be possible to setup the code that generates the probability that a fit is for a certain type of star/sightline to be similar to what is done get compute the 1D pPDFs?  Specifically saving the indexes for each type so that the probability can be computed quickly for a large number of stars.  I think that will be the use case.  Setting it up like this would mean it could be run as part of the fitting *or* as a post-processing step - equally quickly.

karllark:
I see the strength of having each type of star/sightline defined by a function, but this could generate a significant amount of overhead especially when the types we want to compute before large.  Would there not be a way to specify this using a dictionary that specifies the min/max for each BEAST parameter?  And then passing this to a single function that works on said dictionary?   It might even be better if the types could be specified in an ascii table - making it very easy to expand the list of types for which probabilities to be computed.

karllark:
Having some set of the star/sightline probabilities computed during the fit will likely be very useful.  And I'm betting that there is a default set that makes a great deal of sense to do for all BEAST fits.  For example, hot stars, cool stars, young stars, old stars, red giants, lightly extinguished stars, moderately extinguished, heavily extinguished.  Just name a few.

lea-hagen:
> I wonder if this PR should be split into two PRs. One for the tools that merge the lnps from subgrid runs and one to address determining the probabilities of different types of stars/sightlines. @lea-hagen : is this possible? Having two PRs would make it easier to know which changes are for which purpose and make reviewing/accepting the PRs easier.

Definitely possible.  I'm not sure logistically how to do that... maybe close this one, and cherry-pick the relevant commits onto new branches?

------

> I wonder if it would be possible to setup the code that generates the probability that a fit is for a certain type of star/sightline to be similar to what is done get compute the 1D pPDFs? Specifically saving the indexes for each type so that the probability can be computed quickly for a large number of stars. I think that will be the use case. Setting it up like this would mean it could be run as part of the fitting _or_ as a post-processing step - equally quickly.

I'm not quite sure if I'm following the question.  The code does use the whole array of stars (shape of `(n_lnp, n_stars)`), so the operations are relatively optimized.

------

> I see the strength of having each type of star/sightline defined by a function, but this could generate a significant amount of overhead especially when the types we want to compute before large. 

Could you clarify why would this make a lot of overhead?

> Would there not be a way to specify this using a dictionary that specifies the min/max for each BEAST parameter? And then passing this to a single function that works on said dictionary? It might even be better if the types could be specified in an ascii table - making it very easy to expand the list of types for which probabilities to be computed.

I thought about this option, but I couldn't figure out a simple way to arbitrarily define a region.  For instance, I recall that the cuts @s-goldman made for selecting AGB stars are some sort of polygon in two parameters.  If cuts involve something more complicated than min/max, or require more than a couple of parameters, the ascii/dictionary method could get really messy.  However, I'm very much open to changing that if it would be less awful than I'm imagining.

------

> Having some set of the star/sightline probabilities computed during the fit will likely be very useful. And I'm betting that there is a default set that makes a great deal of sense to do for all BEAST fits. For example, hot stars, cool stars, young stars, old stars, red giants, lightly extinguished stars, moderately extinguished, heavily extinguished. Just name a few.

Absolutely.  If we can finalize a set in the next month or so, I can incorporate them into the METAL production runs.

karllark:
> > I wonder if it would be possible to setup the code that generates the probability that a fit is for a certain type of star/sightline to be similar to what is done get compute the 1D pPDFs? Specifically saving the indexes for each type so that the probability can be computed quickly for a large number of stars. I think that will be the use case. Setting it up like this would mean it could be run as part of the fitting _or_ as a post-processing step - equally quickly.
> 
> I'm not quite sure if I'm following the question. The code does use the whole array of stars (shape of `(n_lnp, n_stars)`), so the operations are relatively optimized.

Ok.  That does cut down on the overhead.  But this setup does mean that the code has to be run as a post-processing step.  In other words, it would be hard to efficiently run it like the fitting is done - star-by-star.

karllark:
> > I see the strength of having each type of star/sightline defined by a function, but this could generate a significant amount of overhead especially when the types we want to compute before large.
> 
> Could you clarify why would this make a lot of overhead?

The overhead is that a function needs to be written for all star/sightline types, including the simple ones that could be defined by a simple box.  Maybe there would be a way to have a dictionary for the simple box-like types and then use a function for the others?

lea-hagen:
I'm going to close this PR so its contents can be split into two PRs.

Issue #413, opened 2019-10-17
Spectral grid tests failing astropy dev CI tests


karllark:
Tests failing with astropy development version.  See PR #412 for focused travis CI tests and the failures.

jwuphysics:
I'm working on this-- will see if we can provide an `if` statement for `astropy < 4.0` versions.

PR #471, opened 2020-01-07
Fix edges of SD maps with better area calculation

lea-hagen:
We've had problems with the edges of source density (SD) maps having artificially low values.  This is because only part of the catalog overlaps the map pixel, but the full pixel area is used when evaluating sources/arcsec^2.

I changed the code to create a convex hull around the catalog, and then find the intersection between the convex hull and the pixel.  This provides an area scaling factor, so the pixel area (usually 25 arcsec^2) can be adjusted as needed for any partial overlap.

Here's an example for a METAL field:

| Before | After | 
| --- | --- |
| sd_before | sd_after |

### DIFF
--- a/beast/tools/create_background_density_map.py
+++ b/beast/tools/create_background_density_map.py
@@ -21,9 +21,10 @@ from matplotlib.collections import PatchCollection
 import numpy as np
 import photutils as pu
 from beast.tools.density_map import DensityMap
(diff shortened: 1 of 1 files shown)


karllark:
Crap.  astropy 4.0 used.  Tests fail.  Sigh.  Going to merge as this is a known problem (#413).   Will have to fix this soon.

PR #474, opened 2020-01-17
Added code to convert hdf5 to fits

cmurray-astro:
Added Lea's code to convert hdf5 to fits into tools/ for use in BEAST examples

### DIFF
--- a/beast/plotting/plot_indiv_fit.py
+++ b/beast/plotting/plot_indiv_fit.py
@@ -515,8 +515,8 @@ if __name__ == "__main__":  # pragma: no cover
     filters = [
         "HST_WFC3_F275W",
         "HST_WFC3_F336W",
-        "HST_WFC3_F475W",
-        "HST_WFC3_F814W",
+        "HST_ACS_WFC_F475W",
+        "HST_ACS_WFC_F814W",
         "HST_WFC3_F110W",
         "HST_WFC3_F160W",
     ]
--- a/beast/plotting/plot_indiv_fit.py
+++ b/beast/plotting/plot_indiv_fit.py
@@ -515,8 +515,8 @@ if __name__ == "__main__":  # pragma: no cover
     filters = [
         "HST_WFC3_F275W",
         "HST_WFC3_F336W",
-        "HST_ACS_WFC_F475W",
-        "HST_ACS_WFC_F814W",
+        "HST_WFC3_F475W",
+        "HST_WFC3_F814W",
         "HST_WFC3_F110W",
         "HST_WFC3_F160W",
     ]
--- a/beast/tools/tests/test_convert_hdf5_to_fits.py
+++ b/beast/tools/tests/test_convert_hdf5_to_fits.py
@@ -1,0 +1,12 @@
+from astropy.tests.helper import remote_data
+from beast.tools import convert_hdf5_to_fits
+
+
+@remote_data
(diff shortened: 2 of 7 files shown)


karllark:
Would be great to have a little documentation as part of this PR.

And the two unused imports should be deleted.

cmurray-astro:
Can do! For documentation, do you mean as part of the workflow (in like a "data preparation" step) or should we have a more general documentation of all the tools/helper functions (probably both)?

karllark:
I think we have a place now in the docs for generally listing the tools we have.  But I'm not 100% sure.  We should have such a section.  :-)

lea-hagen:
We have a section that describes plotting tools ([here](https://beast.readthedocs.io/en/latest/plotting_tools.html)), but as far as I can tell, the other things in `beast/tools/` are only described when they come up in the context of something else (the [workflow](https://beast.readthedocs.io/en/latest/workflow.html), info about [generating ASTs](https://beast.readthedocs.io/en/latest/generating_asts.html), etc).

karllark:
Waiting on docs.  :-)

karllark:
@cmurray-astro : can you send me the file(s) I need to add to the cache so that the test can be run by the CI?

beast/tools/tests/test_convert_hdf5_to_fits.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+@remote_data
+def test_convert_hd5_to_fits():
+
+    # Pick some random .hd5 file to convert
+    data_fname = download_rename("beast_example_phat_seds_trim.grid.hd5")
+
+    # Convert the file
+    convert_hd5_to_fits.st_file(data_fname)
```
karllark: It is better practice to import the function needed from `convert_hd4_to_fits` instead of the whole module.  In this case `from convert_hd5_to_fits import st_file` and then this line would be `st_file(data_fname).

Also, this test does not have an assert statement.  Is this intended?  Is this just a "smoke" test to make sure that the conversion happens?  It would be a more complete test if a cached version of the fits file was compared to the just created version to make sure the conversion happened correctly.

cmurray-astro:
Thanks for the comments, Karl! I think I added the right kind (or at least a kind) of comparison between the output file in the test and a cached version (sent email today with those files for caching).

beast/plotting/plot_indiv_fit.py
```diff
@@ -515,8 +515,8 @@ def rectangle_around_axes(bottomleft_ax, topright_ax, pad, ls, label=None):
     filters = [
         "HST_WFC3_F275W",
         "HST_WFC3_F336W",
-        "HST_ACS_WFC_F475W",
-        "HST_ACS_WFC_F814W",
+        "HST_WFC3_F475W",
+        "HST_WFC3_F814W",
```
karllark: Humm.... this changes the default filters for the plot_indiv_fit.  I think this means that the `phat_small` example will not plot.  Was this intentional?

cmurray-astro: Nope! I didn't mean for that to be in this at all! Will investigate

cmurray-astro: Fixed! (I think)

karllark:
Great PR! It has code, tests, and docs. Kudos @cmurray-astro!

PR #479, opened 2020-01-17
simplify 1D PDF code

lea-hagen:
This PR will eventually have code to make 2D PDFs.  The current changes are to make the 1D PDF code more readable (so I can extrapolate it to 2D PDFs).  I ran it on phat_small and it takes the same amount of time to do the fitting.


karllark:
Cleaner code and easier to understand I fully agree. 
I suggest that this PR just be about updating the 1D PDFs (PR name edit needed) and a separate PR be opened for the addition of the 2D PDFs.  Just easier to review.

lea-hagen:
Works for me!  I'll change the PR name and then it should be good to merge.

lea-hagen:
I removed the `gen1d_full` function.  As far as I can tell (using grep), it's not used anywhere.

PR #482, opened 2020-01-17
Update the percentile function to speed up

galaxyumi:
I updated the percentile function in the common.py to speed up the calculation. It returned the same stats results as a run with the original percentile function, with a 14% increase in running time for the PHAT_small example case. I also compared the running time between the expectation value function in the common.py and numpyp.average, and found that our current function is slightly faster than using numpy.average . I left a comment about this in the code.

### DIFF
--- a/beast/fitting/fit_metrics/common.py
+++ b/beast/fitting/fit_metrics/common.py
@@ -158,7 +158,6 @@ def expectation(q, weights=None):
     (1) This function is about 30% fater than usning numpy.average
         to compute expectation values -- test by Yumi Choi on 1/17/2020
     """
-    n = len(q)
 
     if weights is None:
         return np.mean(q)
--- a/beast/fitting/fit_metrics/common.py
+++ b/beast/fitting/fit_metrics/common.py
@@ -44,15 +44,6 @@ Note that the 50th weighted percentile is known as the weighted median.
 """
 import numpy as np
 
-try:
-    from .c_common import weighted_percentile as c_wp
-    from .c_common import expectation as c_expect
-
-    _C_code = True
-except ImportError:
-    _C_code = False
-
-_C_code = False
 
 
 def percentile(data, percentiles, weights=None):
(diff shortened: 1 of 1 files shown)


beast/fitting/fit_metrics/common.py
```diff
@@ -119,42 +119,21 @@ def percentile(data, percentiles, weights=None):
...
-                f2 = (p - w[s - 1]) / (w[s] - w[s - 1])
-                if not (
-                    (f1 >= 0) and (f2 >= 0) and (f1 <= 1) and (f2 <= 1)
-                    and (abs(f1 + f2 - 1.0) < 1e-6)
-                ):
-                    raise AssertionError()
-                o[pk] = sd[s - 1] * f1 + sd[s] * f2
-        return o
+    
+    i = np.argsort(_data)
```
karllark: `i` usually refers to an integer in a loop.  Just for consistency - could you change this to be something more descriptive than `i`?  Like `indxs` or something like that?

karllark: Also, the if statement to the c code has been removed.   I'm find with this, but isn't there some C code in the repository that needs removing as well then?

beast/fitting/fit_metrics/common.py
```diff
@@ -182,6 +161,11 @@ def expectation(q, weights=None):
     -------
     e: float
         expectation value
+    
+    NOTE
+    -------
+    (1) This function is about 30% fater than usning numpy.average 
+        to compute expectation values -- test by Yumi Choi on 1/17/2020
```
karllark: Note much appreciated!

beast/fitting/fit_metrics/common.py
galaxyumi: 'i' has been changed to 'isort'. 

At the beginning of the code, you can see '_C_code = False' in line 55. I assume someone made a decision not to use a function from C to compute the weighted percentile. So I removed the if statement inside the function. If we want to leave the option to use the function from C when is available, I can revert the if statement and remove the line 55.

karllark:
Please remove the unused import identified by LGTM and the issues raised by Coadacy.  Just nice to keep the code is good shape overall.

karllark:
I think we should remove all the references to the C code, both in imports and if statements.  That C code is no longer present in the current version of the repository and so it is not possible to ever use it (without going to previous versions and getting it back).  Best to just remove it all - more consistent and clean.

galaxyumi:
> Please remove the unused import identified by LGTM and the issues raised by Coadacy. Just nice to keep the code is good shape overall.

karllark:
Unused variable identified by LGTM needs to be removed.

PR #491, opened 2020-01-22
Remove logL=-9.999 points in the spec grid stage not to allow them in the final SED grid

### LINKED ISSUE #486: log(L) = -9.999 model points in the isochrones
There are some logL = -9 models in the PARSEC isochrones, and it is unclear what they are. I emailed Leo Girardi about this. 
logL_minus9

galaxyumi:
The model points with logL = -9.999 (see issue #486) has been removed before generating the spec grid. 

Closes #486


karllark:
Looks great.  Thanks.

PR #493, opened 2020-01-23
Merging lnPs and file reading tools

lea-hagen:
This is basically a duplicate of #468, but with just the tools in that bulleted list, and not the star type probability code.  I've copied over that list here:

* `tools/subgridding_tools.py`: Added a `merge_lnp` function to combine lnPs across subgrids.  There's an optional `threshold` argument (like the one in `fit.py`) to eliminate lnPs much smaller than the maximum, which can help reduce file size.
* `tools/run/merge_files.py`: includes `merge_lnp` in the merging for subgrids
* `tools/read_beast_data.py`: Migrated over from `read_beast.py` in the MegaBEAST (BEAST-Fitting/megabeast#40).  It has functions to extract data from the files for the noise model, SED grid, and lnPs, and return them as dictionaries.  For the SED grid, if the parameter list is `None`, it will print out all of the parameters in the file (relevant to #433).  The code also has a function to get the parameter values associated with the lnPs.
* `tools/trim_many_via_obsdata.py`: If a trimmed file already exists, it will now skip to the next one in the input file.  (This isn't immediately related, but it came up while I was testing a start-to-finish run with the lnP merging.)

### DIFF
--- a/beast/tools/read_beast_data.py
+++ b/beast/tools/read_beast_data.py
@@ -99,14 +99,14 @@ def read_noise_data(
     noise_data = {}
 
     # open files for reading
-    with h5py.File(noise_filename, 'r') as noise_hdf:
+    with h5py.File(filename, 'r') as noise_hdf:
 
         # get beast physicsmodel params
         for param in tqdm(param_list, desc='reading beast data'):
             if filter_col is None:
-                noise_data[cparam] = np.array(noise_hdf[cparam])
+                noise_data[param] = np.array(noise_hdf[param])
             else:
-                noise_data[cparam] = noise_hdf[cparam][:,filter_col]
+                noise_data[param] = noise_hdf[cparam][:,filter_col]
 
     return noise_data
 
@@ -190,9 +190,9 @@ def get_lnp_grid_vals(sed_data, lnp_data):
         arrays of the SED grid parameters for the points in the lnp lists
     """
 
-    if type(sed_data) == str:
+    if isinstance(sed_data, str):
         sed_data = read_sed_data(sed_data)
-    if type(lnp_data) == str:
+    if isinstance(lnp_data, str):
         lnp_data = read_lnp_data(lnp_data)
 
     # get the keys in beast_data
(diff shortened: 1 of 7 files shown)


karllark:
There is no update to the docs for these changes.  Is it because everything is already documented?  I feel like this PR has some changes that are not reflected in the current docs.  Please advise.

lea-hagen:
I added some documentation about the new lnP merging tool and the reading tools.  Is there anything else I should address before this is merged?

beast/tools/read_beast_data.py
```diff
@@ -0,0 +1,215 @@
+"""
+Functions for interacting with the BEAST model
+"""
+
+# system imports
+from __future__ import (absolute_import, division, print_function)
```
karllark: We don't support python 2, so the future imports can be removed.

beast/tools/run/merge_files.py
```diff
@@ -5,8 +5,10 @@
 from beast.tools.run import create_filenames
 
 
+
 import datamodel
 import importlib
+importlib.reload(subgridding_tools)
```
karllark: Remove.

beast/tools/trim_many_via_obsdata.py
```diff
@@ -60,6 +60,11 @@
         sed_trimname = filebase + "_seds_trim.grid.hd5"
         noisemodel_trimname = filebase + "_noisemodel_trim.grid.hd5"
 
+        # if these already exist, then continue to the next set of files to trim
+        if os.path.isfile(sed_trimname) and os.path.isfile(noisemodel_trimname):
+            print("trimming already complete for "+sed_trimname)
+            continue
+
```
karllark: This is not related to the main topic of this PR.  Could be a separate PR.

Not sure that not just automatically overwriting the trim files isn't a better solution.  That is what would currently happen, right?  Or is the current behavior to crash?

lea-hagen: Happy to move it to a new PR.

The issue is that trimming can take a while, so skipping the file makes it run faster.

karllark: Please do.  I mainly want to make sure that this change does not mess up runs as right now I'm assuming that is does replace the file.  Would be a bummer if this silently changed to *not* updating the file and this was not the desired behavior.

lea-hagen:
Just did a force push to undo merging master into this branch.  I'll do a rebase on my computer to resolve the documentation conflicts.

beast/tools/trim_many_via_obsdata.py
lea-hagen: Opened a new PR for this (#501).

Issue #472, opened 2020-01-16
Saving 2D PDFs


lea-hagen:
We've been trying to solve #373 (evaluating the probability of different stellar types) using the log likelihoods - either calculating probabilities during the BEAST run, or saving enough to evaluate afterwards.  Both have significant disadvantages (see discussions in #373, #467, #468).  

Another option I'm investigating is saving 2D PDFs.  It'll require a fair bit of disk space, so if we decide to go this route, we should decide what quantities are worth saving.  To start, here's a list of parameters (from one of the METAL fields) that get 1D PDFs:
```
['Av',
 'M_act',
 'M_ini',
 'Rv',
 'Rv_A',
 'Z',
 'distance',
 'f_A',
 'logA',
 'logHST_ACS_WFC_F475W_nd',
 'logHST_ACS_WFC_F475W_wd',
 'logHST_ACS_WFC_F814W_nd',
 'logHST_ACS_WFC_F814W_wd',
 'logHST_WFC3_F110W_nd',
 'logHST_WFC3_F110W_wd',
 'logHST_WFC3_F160W_nd',
 'logHST_WFC3_F160W_wd',
 'logHST_WFC3_F225W_nd',
 'logHST_WFC3_F225W_wd',
 'logHST_WFC3_F275W_nd',
 'logHST_WFC3_F275W_wd',
 'logHST_WFC3_F336W_nd',
 'logHST_WFC3_F336W_wd',
 'logL',
 'logT',
 'logg',
 'mbolmag',
 'radius']
```
I don't think we'd need any of the flux-related quantities.  Among the other physical parameters, are there any that we can definitely ignore, or any pairs that are unlikely to ever be useful?

karllark:
In helping decide which 2D pPDFs to save, it would be useful to know what these will be used for.  What is the science goal/question for which such information is needed?

lea-hagen:
The idea is to be able to make cuts in two parameters to evaluate the probability of a particular stellar type.  For instance, if you wanted the probability that your source is an O star, you could integrate the `M_ini` 1D PDF between the relevant masses, and normalize by the total.  But if you wanted the probability that it's a good candidate for extinction curves, you'd need to integrate within the `M_ini - A_V` 2D PDF.

karllark:
Ok.  You've made an argument for 1 2D pPDF.  Are there others we need for similar reasons?  Or are you thinking to make the *all* just in case?  Just trying to better understand the question.  :-)

[...]

PR #494, opened 2020-01-23
Star type probability

lea-hagen:
Along with #493, this recreates the content in #468.  This PR has just the code to create star type probabilities from the saved lnPs.  However, since we've decided to try going the route of 2D PDFs rather than using lnPs (#472), I'll need to edit this.

### DIFF
--- a/beast/tools/star_type_probability.py
+++ b/beast/tools/star_type_probability.py
@@ -1,8 +1,5 @@
 import numpy as np
 from collections import defaultdict
-import matplotlib.pyplot as plt
-
-from beast.tools import read_beast_data
 
 from astropy.table import Table
 from astropy.io import fits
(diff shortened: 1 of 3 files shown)


lea-hagen:
The code/docs are complete.  Feedback (small or large) appreciated, particularly on the idea of making some sort of ascii input list of cuts (as @karllark suggested in #468).

karllark:
No longer WIP?  Please remove WIP from the title of this PR.

beast/tools/star_type_probability.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def star_type_probability(
+    pdf1d_files,
+    pdf2d_files,
+    output_filebase=None,
+    ext_O_star_params=None,
+    dusty_agb_params=None,
+):
+    """
```
karllark: Need a short description of what this function is doing.

karllark: Standard part of docstrings.

lea-hagen: Thanks for catching that!

PR #495, opened 2020-01-23
Make 2D PDFs

### LINKED ISSUE #472: Saving 2D PDFs

lea-hagen:
This is most of the work needed to create 2D PDFs.  The only missing thing is merging across source densities and/or subgrids, but I think that'll be best left to a different PR after #493 is merged.

Here's what I've changed/added:
* `fitting/pdf2d.py`: A new class (very similar to `pdf1d.py`) that assigns grid indices to each combination of the two input parameters, which later can be used to quickly evaluate the marginalized probabilities.
* `fitting/fit.py`:
  * `setup_param_bins`: Moved code that generates 1D PDF parameter bins over to its own function, since otherwise it would've been basically copy-pasted for the 2D PDFs.
  * `save_pdf2d`: saves the 2D PDFs to a fits file.  The format is one extension for each pair of parameters.
  * added code that creates the fast 2D PDF objects and uses them to calculate marginalized 2D PDFs
  * added `pdf2d_outname` as an optional input for `summary_table_memory` and `Q_all_memory`
* running tools (`tools/run/`)
  * `create_filenames.py`: added the 2D PDF filenames to the dictionary of file names it creates
  * `run_fitting.py`: update to create 2D PDFs by default (I'm assuming that's what we want?)
* `plotting/plot_indiv_pdfs.py`: makes a triangle plot of the 2D and 1D PDFs.  Examples for coarse grids are below.  The 1D PDFs are linear, and the 2D PDFs are log-scaled (is that confusing?).  There's also a section that will plot 1/2/3-sigma contours on the 2D PDFs if there are enough grid points to distinguish them.

Closes #472

| phat_small star 0 | a METAL star |
| --- | --- |
| phat_small_pdfs_0 | metal_pdfs_20 |

### DIFF
--- a/beast/tools/run/run_fitting.py
+++ b/beast/tools/run/run_fitting.py
@@ -26,6 +26,7 @@ def run_fitting(
     nprocs=1,
     choose_sd_sub=None,
     choose_subgrid=None,
+    pdf2d_param_list=['Av', 'Rv', 'f_A', 'M_ini', 'logA', 'Z', 'distance'],
     resume=False,
 ):
     """
@@ -52,11 +53,14 @@ def run_fitting(
     choose_sd_sub : list of two strings (default=None)
         If this is set, the fitting will just be for this combo of SD+sub,
         rather than all of them.  Overrides use_sd.
-        format of the list: ['#-#','#']
+        format of the list: ['#','#']
 
     choose_subgrid : int (default=None)
         If this is set, the fitting with just be for this subgrid index.
         If nsubs=1, this is ignored.
+
+    pdf2d_param_list : list of strings or None
+        If set, do 2D PDFs of these parameters.  If None, don't make 2D PDFs.
 
     resume : boolean (default=False)
         choose whether to resume existing run or start over
@@ -94,6 +98,8 @@ def run_fitting(
     stats_files = file_dict["stats_files"]
     pdf_files = file_dict["pdf_files"]
     pdf2d_files = file_dict["pdf2d_files"]
+    if pdf2d_param_list is None:
+        pdf2d_files = [None for i in range(len(pdf2d_files))]
     lnp_files = file_dict["lnp_files"]
 
     # total number of files
@@ -158,6 +164,7 @@ def run_fitting(
                 stats_files[i],
                 pdf_files[i],
                 pdf2d_files[i],
+                pdf2d_param_list,
                 lnp_files[i],
                 None,
                 resume,
@@ -175,6 +182,7 @@ def run_fitting(
                 stats_files[i],
                 pdf_files[i],
                 pdf2d_files[i],
+                pdf2d_param_list,
                 lnp_files[i],
                 gridpickle_files[i],
                 resume,
@@ -198,6 +206,7 @@ def fit_submodel(
     stats_file,
     pdf_file,
     pdf2d_file,
+    pdf2d_param_list,
     lnp_file,
     grid_info_file=None,
     resume=False,
@@ -224,6 +233,9 @@ def fit_submodel(
 
     pdf2d_file : string
         path+name of the file to contain 2D PDF output
+
+    pdf2d_param_list: list of strings or None
+        parameters for which to make 2D PDFs (or None)
 
     lnp_file : string
         path+name of the file to contain log likelihood output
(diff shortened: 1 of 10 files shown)


karllark:
Great to see.  Will be really cool to see the 2D pdfs for the fits.

Do you have an estimate of how much this adds to the computation time of a run?  Useful to know for deciding if I want to have 2D pPDFs.  :-)

What about just having a single file for all the pdf material instead of pdf1d and pdf2d?  Maybe something like fastpdf?  Then all the pdf material will be in a single file.

Docs?  How does one use this new capability?  How does one pick which set of 2D pdfs to do?  How to turn them off if one doesn't want them?

lea-hagen:
> Do you have an estimate of how much this adds to the computation time of a run? Useful to know for deciding if I want to have 2D pPDFs. :-)

I just did some runs of the the phat_small example, it takes me 18 seconds with the 2D PDFs and 17 seconds without (those are the times in the `Lnp/Stats` progress bar).  So my initial estimate is that it doesn't add much.  Though I also ran a test with everything (1D/2D PDFs, lnP, stats) turned off, and it still took 15 seconds, so I'm not sure how reliable my estimate is.  I don't have time today, but later I'll check for one of the METAL fields.

> What about just having a single file for all the pdf material instead of pdf1d and pdf2d? Maybe something like fastpdf? Then all the pdf material will be in a single file.

It's nice to have them as separate files because then all fits extensions within a file have the same naming system and data structure.  But it wouldn't be too hard to change it if one file is generally preferred.

> Docs? How does one use this new capability? 

Yeah, I thought of docs right after clicking the button.  On the to-do list :)

> How does one pick which set of 2D pdfs to do? 

It's currently hard-coded to do `['Av', 'Rv', 'f_A', 'M_ini', 'logA', 'Z', 'distance']` based on https://github.com/BEAST-Fitting/beast/issues/472#issuecomment-575326251.

> How to turn them off if one doesn't want them?

Since they apparently don't add much overhead, I would advocate for always leaving them on.  But it's easy to add it as an option in `run_fitting.py` if that would be useful.

lea-hagen:
I updated the output files documentation to include info about the 2D PDFs (as well as the 1D PDFs).  I also reformatted the page a bit to improve the flow.  These all partially address #499.

lea-hagen:
I think I'm misunderstanding something about the docs.  I have a python code block that's intended to show how to access the files, but it seems like the Travis testing is actually looking for that file.  When I build the docs locally, that doesn't happen.  @karllark, do you know what's going on?

lea-hagen:
I was doing some more timing comparisons and realized that the 2D PDFs were being created even if they weren't being saved.  So I've fixed that!

Now for the timing: I ran a coarse METAL grid, and found that doing the 2D PDFs increased the fitting time by 10%.

lea-hagen:
Solved the testing problem - turns out I needed to add `#doctest: +SKIP` in the code blocks.

lea-hagen:
I have added an update to allow the user to choose any of the parameters (not just the seven main ones), or to turn off 2D PDFs altogether.  The docs and plotting code are edited accordingly.

beast/tools/run/run_fitting.py
```diff
@@ -17,6 +17,7 @@
 
 import datamodel
 import importlib
+importlib.reload(fit)
```
karllark: Please remove.

lea-hagen: done!

PR #503, opened 2020-02-03
small change to mag_cut in AST-generating code

lea-hagen:
I noticed that `make_ast_input_list.pick_models_toothpick_style`, which does the flux bin AST selection method, had `mag_cuts` as an input even though the code doesn't use it, so I removed it.  I also moved the chunk of code in `tools/run/make_ast_inputs.py` that constructs `mag_cuts` so that it only runs if the random SED selection method is used.

### DIFF
--- a/beast/observationmodel/ast/make_ast_input_list.py
+++ b/beast/observationmodel/ast/make_ast_input_list.py
@@ -59,7 +59,6 @@ def mag_limits(seds, faint_cut, Nfilter=1, bright_cut=None):
 def pick_models_toothpick_style(
     sedgrid_fname,
     filters,
-    mag_cuts,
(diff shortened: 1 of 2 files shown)


karllark:
Great.  Good catch.

galaxyumi:
Oh, no. Could you revert this pull request? I am actually using/testing the magnitude cuts even for the toothpick style SED sampling to remove SEDs that would be 0 completeness for sure in any bands. Does it make sense?

karllark:
I don't know how to do this.  How about you just put it back in the PR for that will address the work you are currently doing.  Thanks!

galaxyumi:
Okay. I will do a PR on this.

PR #506, opened 2020-02-11
compare BEAST results to spectrally typed stars

lea-hagen:
This code is part of the BEAST verification: comparing spectrally typed stars to the BEAST results.  Here's what it currently does:
* Convert the stars' spectral types to T_eff and log(g) by interpolating on reference tables (see below)
* Find the corresponding star in the BEAST catalog: given that spectrally typed stars tend to be bright, find the brightest BEAST star within a user-defined radius
* Calculate the number of standard deviations the "true" value is from the BEAST value, using p16/p50/p84

To fully test this, I need to find a field with BEAST fits that also has spectrally typed star(s).  I don't have any matches in METAL yet, so suggestions welcomed.

Also need to add docs.

---

The references for the T_eff and log(g) conversions:
* T_eff: Table B.3 and B.4 in Stellar Spectral Classification
* log(g): Table 15.8 in Allen's Astrophysical Quantities

Screen shots of the tables are below.  I'm not sure where to put them - maybe in the library files that the user downloads?  They're not large, but I'm not sure if they should be cluttering up `beast/tools/`.

| T_eff table | log(g) table |
| --- | --- |
|  teff | logg |


lea-hagen:
This has finally been tested with actual stars!  As part of writing my HST proposal to get STIS spectra of BEAST-identified extinguished OB stars, I found that 7 of the [Bresolin+07](https://ui.adsabs.harvard.edu/abs/2007ApJ...671.2028B/abstract) stars are in the IC1613 LUVIT field.  I used this code to convert the spectral types to T_eff/log(g) and compare them to the BEAST fits.  Turns out it does pretty well!

ic1613_spec_compare

For those who are curious, here are the spectral types and numbers:
```
spec_types spec_teff beast_teff_p50 spec_logg beast_logg_p50
---------- --------- -------------- --------- --------------
   A0 II        9950           8913      2.76           1.87
   A5 II        8050           9484      2.65           1.97
   B8 I        11100          12157      2.26           2.47
   B8 I        11100          12157      2.26           2.28
   O9.5 I      30463          43418      2.94           3.88
   A5 I         8100           4231      2.04           1.25
   B5 II       14200          13763      2.96           2.58
```

lea-hagen:
Before this is merged, we need to find a place for the T_eff and log(g) tables to live so that the path in the code is correct.  (Right now it's just pointing to one of my local files.)

karllark:
You can put those files in a data subdirectory of tools.  Then you can just get them by using the installed location of the beast.  Check out the code in dust_extinction referenced below for an example.

https://github.com/karllark/dust_extinction/blob/0734a3042aca7e575aaeca32daa01e8e04abcf91/dust_extinction/parameter_averages.py#L1209

beast/tools/compare_spec_type.py
```diff
@@ -0,0 +1,337 @@
...
+    Convert the luminosity class from roman numeral to integer
+    """
+
+    return ['I','II','III','IV','V'].index(lumin_class) + 1
+
+
+if __name__ == '__main__':
+
+    print('moose')
+    # argparse stuff
```
karllark: This is not needed, right?

lea-hagen: Hahaha, nope :)

lea-hagen:
I think this is good to go now.

PR #460, opened 2019-11-21
Replace sys.exit() calls with informative Warning/Error messages

jwuphysics:
Fixes #427 by removing nearly all `exit()` calls. Instead, I use relevant + more "graceful" crashes, e.g., `NotImplementedError`, `ValueError`, `TypeError`, or in some cases, `UserWarning`. Some of the `exit()` commands are left in because it wasn't clear how to reformulate them (e.g., in `docs/conf.py`.

Note that the original PR (#453) is also included here because I accidentally branched my branch.

PR #507, opened 2020-02-19
Fix warnings in verify_params

lcjohnso:
In #460, @jwuphysics swapped `exit()` interrupts with errors and warnings.  The non-interrupting warnings, which were previously handled by the `noexit` option, were not actually implemented as warnings but as interrupting errors.

This PR swaps the non-existent `UserWarning` with a `warnings.warn()` call (whose default type is UserWarning).

### DIFF
--- a/beast/tools/tests/test_verify_warnings.py
+++ b/beast/tools/tests/test_verify_warnings.py
@@ -1,0 +1,46 @@
+import pytest
+from beast.tools import verify_params
+
+class datamodel_mock:
+    """A simple mock datamodel"""
+    obsfile = 'photometry.fits'
+    astfile = 'ast.fits'
+    logt = [6.0, 10.13, 0.05]
+    age_prior_model = {'name': 'flat_log'}
+    mass_prior_model = {'name': 'kroupa'}
+    z = [0.00452, 0.00254, 0.00143]
+    met_prior_model = {'name': 'flat'}
+    avs = [0.0, 4.0, 0.1]
+    av_prior_model = {'name': 'lognormal',
+                      'max_pos': 0.76,
+                      'sigma': 0.73,
+                      'N': 10.}
+    rvs = [3.0, 6.0, 1.0]
+    rv_prior_model = {'name': 'flat'}
+    fAs = [0.0, 1.0, 0.25]
+    fA_prior_model = {'name': 'flat'}
+
+class datamodel_mock_nofA(datamodel_mock):
+    """Mock datamodel w/ fAs=None"""
+    fAs = None
+
+class datamodel_mock_allowwarn(datamodel_mock_nofA):
+    """Mock datamodel w/ fAs=None and allow_warnings = True"""
+    allow_warnings = True
+
+def test_verifyparams_nowarning():
+    """Test: verify_params for case of no warnings or exceptions."""
+    with pytest.warns(None) as record:
+        verify_params.verify_input_format(datamodel_mock())
+    assert len(record) == 0
+
+def test_verifyparams_error():
+    """Test: verify_params for case of warning raising exception."""
+    with pytest.raises(UserWarning) as exc:
+        verify_params.verify_input_format(datamodel_mock_nofA())
+    assert exc.value.args[0] == "fAs is not defined."
+
+def test_verifyparams_allowwarn():
+    """Test: verify_params for case of warning with no exception."""
+    with pytest.warns(UserWarning, match="fAs is not defined."):
+        verify_params.verify_input_format(datamodel_mock_allowwarn())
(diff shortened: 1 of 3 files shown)


jwuphysics:
Looks good!

karllark:
One question, will the new behavior be to issue a warning or an error?  In other words, does execution continue or stop?  I'm asking as it sounds like execution will now continue and this would be a change from the previous behavior.  Just want to make sure I understand the change fully.

jwuphysics:
Ah good point, sorry I totally missed that! `raise` will exit the program (similar to the original behavior) but `warnings.warn()` will continue. So in fact the behavior will change.

lcjohnso:
The change I'm trying to reverse is the loss of the `noexit` option -- an ignorable, non-interrupting version of these warnings was removed.  As valid cases exists where `fA` is not used by the extinction model, or grid params are single valued, the current forced behavior of throwing an interrupting error is not acceptable.

It seems like "best of both worlds" would be that the current warnings throw an interrupting error by default, but that they can be set to ignorable.  That can be accomplished by using `warnings.simplefilter('error', UserWarning)` by default, which turns warnings into errors, but provide an `allow_warnings` option that would skip the `simplefilter` call and make the warnings  non-interrupting.  Essentially reverting the loss of the `noexit`-like option that was removed by #460...

@karllark & @jwuphysics: if this is a reasonable compromise, I'll implement this.

karllark:
This sounds like a great solution.  Go for it.

jwuphysics:
I agree with Karl. Thanks for catching the change in behavior.

lcjohnso:
New implementation is in place: default behavior is to raise exception, but if `allow_warnings` is set to `True` in datamodel.py, then warnings become non-interrupting.

I need to add a test: if `fA=None` in datamodel.py, then raise UserWarning exception during `verify_parameters.verify_input_format()` by default.  If `datamodel.allow_warnings` is `True`, then raise non-interrupting warning during `verify_parameters.verify_input_format()` and continue.

Help on implementing this test would be much appreciated!

karllark:
A test would be awesome!

Have you looked at the existing tests for examples?  Or are they not sufficient?  There are examples of tests that look for warnings/errors in the dust_extinction package if more examples would be useful.  I can also just dig out probably what you need.

lcjohnso:
I'm stuck on writing a test for this.  My idea was to model a couple tests from [dust_extinction/tests/test_warnings.py](https://github.com/karllark/dust_extinction/blob/master/dust_extinction/tests/test_warnings.py), but the fact that verify_params.verify_input_format() takes `datamodel` module as input means I'd really need to create test versions (e.g., datamodel_test_noallow.py, datamodel_test_allow.py), either in beast/tools/tests/ or externally, that could be imported.  Yuck.

In the meantime, I found and corrected a bug by feeding verify_params.verify_input_format() various test datamodels -- I wish there was an easier way to capture that process in a coded test.  Suggestions welcome!

karllark:
datamodel.py could be mocked up as a class.  This is how it behaves.  Could this solve  the problem?

lcjohnso:
Yes, I think that would do it.  I'll give this a shot and ping again when I've got code to review.

lcjohnso:
@karllark Tests are in place and this PR is ready for review and merge.  Looks like the Codacy CI tool should be tuned so it doesn't complain about `assert` calls in tests.

beast/tools/tests/test_verify_warnings.py
```diff
@@ -0,0 +1,76 @@
...
+    """Mock datamodel w/ single-valued R_V and allow_warnings = True"""
+    allow_warnings = True
+
+def test_verifyparams_errorRV():
+    """Test: verify_params for case of warning raising exception."""
+    with pytest.raises(UserWarning) as exc:
+        verify_params.verify_input_format(datamodel_mock_RV())
+    assert exc.value.args[0] == "Note: rvs grid is single-valued."
+
+def test_verifyparams_errorRV():
```
karllark: Duplicate function name.  Should be different.  Or is this just a duplication of the entire function?

karllark:
Working on removing the codacy check of assert statements.  I agree this is annoying.

beast/tools/tests/test_verify_warnings.py
jwuphysics: Duplicate function name: the two functions differ as one verifies `datamodel_mock_RV()` and the other `datamodel_mock_noallowRV()`.

lcjohnso: Thanks for catching.  This was me wanting to add one more case to test; I'll edit function name and docstring, so should be good to go.

lcjohnso:
> Working on removing the codacy check of assert statements. I agree this is annoying.

Great -- would rather have fewer CI formatting checks than start ignoring PR check failures.

karllark:
Great PR!  It has code, tests, and docs.  Kudos @lcjohnso!

PR #511, opened 2020-02-21
Fix string search for identifying bin#/sub# in file names

lea-hagen:
There are a few places where the bin number and sub-bin number of a file are found by locating "_bin" and "sub" in the file name.  These were done with `.find()`, so if the file name had those sets of characters elsewhere, it would incorrectly use those positions.  I switched to `.rfind()`, since the bin/sub info will always be at the end of the file name.

Thanks @christinawlindberg for finding the bug!

This needs to be included in the production run branch.


karllark:
Looks good.  Merging.

PR #513, opened 2020-02-26
read_beast_data.py: option to read in whole SED grid

lea-hagen:
In the `read_sed_data` function of `read_beast_data.py`, the only options were to specify a list of parameters to read or have it return the whole list of available parameters.  Now it will also accept `param_list='all'` to read in the entire file.

### DIFF
--- a/beast/tools/read_beast_data.py
+++ b/beast/tools/read_beast_data.py
@@ -124,6 +124,7 @@ def read_sed_data(
        the set of parameters to extract
        default = [Av, Rv, f_A, M_ini, logA, Z, distance]
        If set to None, return the list of possible parameters
+       If set to 'all', read in all parameters
 
     Returns
     -------
@@ -143,6 +144,8 @@ def read_sed_data(
         # return that if the user is so inclined
         if param_list is None:
             return grid_param_list + ['seds', 'lamb']
+        if param_list == 'all':
+            param_list = grid_param_list
 
         # get parameters
         for param in tqdm(param_list, desc='reading beast data'):
--- a/beast/tools/read_beast_data.py
+++ b/beast/tools/read_beast_data.py
@@ -111,6 +111,7 @@ def read_noise_data(
 def read_sed_data(
     filename,
     param_list=['Av', 'Rv', 'f_A', 'M_ini', 'logA', 'Z', 'distance'],
+    return_params=False,
 ):
     """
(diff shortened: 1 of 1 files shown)


beast/tools/read_beast_data.py
```diff
@@ -124,6 +124,7 @@ def read_sed_data(
        the set of parameters to extract
        default = [Av, Rv, f_A, M_ini, logA, Z, distance]
        If set to None, return the list of possible parameters
+       If set to 'all', read in all parameters
```
karllark: Need a bit more documentation on how `None` is different from `all`.

lea-hagen: Yeah, that was a bit confusing.  I made a separate keyword for returning the list of parameters, so hopefully that's more clear.

lea-hagen:
Assuming no additional edits, this is ready to merge.

beast/tools/read_beast_data.py
```diff
@@ -111,19 +111,23 @@ def read_noise_data(
...
     param_list : list of strings
-       the set of parameters to extract
-       default = [Av, Rv, f_A, M_ini, logA, Z, distance]
-       If set to None, return the list of possible parameters
+        The set of parameters to extract (default: Av, Rv, f_A, M_ini, logA, Z,
+        distance).  If set to 'all', extract all parameters in the grid.
+
+    return_params : boolean (default=False)
+        If True, return the list of all parameters in the grid.  Useful for
+        checking what columns are present.
```
karllark: This is still not clear to me.  If this keyword is set, then all the parameters *and* the seds and lams are returned.  This sounds like more than just all the parameters.

lea-hagen: `seds` and `lamb` are also stored in the model grid, they're just in different branches (is that the right word?) of the hdf5 file.  So it was easier to manually add them to the list of parameters, rather than formulaically add them in the same way as the other parameters.

karllark: But they are not parameters.  So stating that all parameters are returned is a bit misleading as *more* than all the parameters are returned.

lea-hagen: Ah, I think I understand now.  Would a better phrasing be "parameters and model fluxes"?

karllark: Yes.

PR #516, opened 2020-02-28
distance grid and prior weights (fixes multidistance bug)

### LINKED ISSUE #45: Adjustable distance prior
There is no distance prior.  Issue #38 will add a way to have distance as a fit parameter.  But with a flat prior.  We should have an adjustable distance prior.

Better would be to have the form and necessary variables set in the datamodel.py file or something else. Something else may be better for eventual use by the megabeast.

Two motivations.

Provide flexibility for the input of the distance prior for BEAST runs
Provide flexibility for the eventual use of the BEAST distance model by the MegaBEAST

### LINKED ISSUE #514: physics model prior weights are NaN
I just generated a physics model grid using the standard procedure, and all of the values in `weight` and `prior_weight` are NaN.  I'm not sure if it's something weird I did or a bug that recently showed up, but I'm currently investigating.  If someone else has a recently-generated grid and could take a quick look, that would also be helpful.

karllark:
Fixes the bug that resulted in NaNs for the grid weights when more than 1 distance was present in the grid. Closes #514.

As part of this fix, it made sense to implement the code to explicitly compute the grid and prior weights for distance.  Done similar to how the same for metallicity is handled.  Like metallicity, a flat prior is the only currently supported prior shape.  Other distance priors can be easily implemented now.  Closes #45.

### DIFF
--- a/beast/physicsmodel/grid_and_prior_weights.py
+++ b/beast/physicsmodel/grid_and_prior_weights.py
@@ -14,19 +14,100 @@ with prior weights.
 """
 import numpy as np
 
+from beast.physicsmodel.grid_weights_stars import compute_distance_grid_weights
 from beast.physicsmodel.grid_weights_stars import compute_age_grid_weights
 from beast.physicsmodel.grid_weights_stars import compute_mass_grid_weights
 from beast.physicsmodel.grid_weights_stars import compute_metallicity_grid_weights
 
+from beast.physicsmodel.prior_weights_stars import compute_distance_prior_weights
 from beast.physicsmodel.prior_weights_stars import compute_age_prior_weights
 from beast.physicsmodel.prior_weights_stars import compute_mass_prior_weights
 from beast.physicsmodel.prior_weights_stars import compute_metallicity_prior_weights
 
-__all__ = ["compute_age_mass_metallicity_weights"]
+__all__ = ["compute_age_mass_metallicity_weights",
+           "compute_distance_age_mass_metallicity_weights"]
+
+
+def compute_distance_age_mass_metallicity_weights(
+    _tgrid,
+    distance_prior_model={"name": "flat"},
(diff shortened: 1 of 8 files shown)


lea-hagen:
Alright, finally tested this!  (As a side note, github even suggests how to fetch/check out the branch locally so I don't have to clone it!)  The whole BEAST-ing process worked well, and all the weights appear to be reasonable.  Thanks for tackling these fixes!

A few thoughts/questions:
* At some point, we had talked about how grid weights will differ if distances are in length vs. magnitudes.  I *think* this isn't actually an issue, since the distances are always converted to parsecs before calculating grid/prior weights.
* Could you add `distance_prior_model = datamodel.distance_prior_model` to line 110 in `tools/run/create_physicsmodel.py`?
* Could you update the `datamodel.py` files in beast-examples to include the distance prior?

karllark:
> Alright, finally tested this! (As a side note, github even suggests how to fetch/check out the branch locally so I don't have to clone it!) The whole BEAST-ing process worked well, and all the weights appear to be reasonable. Thanks for tackling these fixes!

You are welcome.  Thanks for finding the bug and reporting it as an issue.

> 
> A few thoughts/questions:
> 
> * At some point, we had talked about how grid weights will differ if distances are in length vs. magnitudes.  I _think_ this isn't actually an issue, since the distances are always converted to parsecs before calculating grid/prior weights.

As long as the distances are converted to parsecs before calculating, this is not an issue.

> * Could you add `distance_prior_model = datamodel.distance_prior_model` to line 110 in `tools/run/create_physicsmodel.py`?

Done.

> * Could you update the `datamodel.py` files in beast-examples to include the distance prior?

Will do in a separate pull request.

karllark:
@lea-hagen : please officially approve this PR if you are happy with it.

PR #518, opened 2020-02-29
Changes for more high mass stars in simulated catalogs

### LINKED ISSUE #517: Issues with creating catalogs for BEAST verification
I'm doing a verification run in which I use a model/noise grids to create a simulated catalog.  That catalog has very few, if any, high-mass stars (which I'm defining as >3 Msun).  I've narrowed it down to two related reasons.

1. Mass prior: The code in `gen_SimObs_from_sedgrid` weights the likelihood of a given SED point by its weight, which is a combination of grid and prior weights.  The prior for mass is the IMF, which means it preferentially selects low-mass stars.  I did a test in which I change it to use just the grid weights, which is a flat mass prior, but I still have the same problem.

2. Age prior: Because the age grid is log-spaced, the grid weighting prefers old ages.  So using the grid weight in `gen_SimObs_from_sedgrid` means picking more low-mass stars (which are still alive).  The prior weight is relatively constant for each age, but if I can't use that without running into the IMF issue.

The one option I've thought of is to ignore the weight info in the SED grid and create my own weights.  Presumably this would just be grid weights for metallicity and mass, since those are the two quantities that could have uneven spacing, and no prior weights.  I haven't done any looking at the grid/prior weight code for the dust parameters, so I'm not sure how this would effect those, particularly Rv/fA, which have a complicated shape.  Perhaps they need some custom weights, too, to make sure that they are fully sampled.

lea-hagen:
As described in #517, the simulations weren't making high mass stars.  I've made two changes to improve this: 
* The code to create simulated observations (`tools/simulate_obs.py`, `gen_SimObs_from_sedgrid` in `observationmodel/observations.py`) now had a keyword `use_weight`.  The user can choose `weight` (prior+grid), `weight_prior`, or `weight_grid` to choose the weighting for selecting of SEDs.
* I added a flat mass prior, so the high mass stars are chosen at more similar rates to low mass stars.

I also updated the docs to describe these.

Closes #517

### DIFF
--- a/beast/observationmodel/observations.py
+++ b/beast/observationmodel/observations.py
@@ -160,6 +160,7 @@ def gen_SimObs_from_sedgrid(
     compl_filter="F475W",
     ranseed=None,
     vega_fname=None,
+    use_weight='weight',
 ):
     """
     Generate simulated observations using the physics and observation grids.
@@ -193,6 +194,10 @@ def gen_SimObs_from_sedgrid(
         filename for the vega info
         usefule for testing
 
+    use_weight : string (default='weight')
+        Set to either 'weight' (prior+grid), 'prior_weight', or 'grid_weight' to
+        choose the weighting for SED selection.
+
     Returns
     -------
     simtable : astropy Table
@@ -224,7 +229,7 @@ def gen_SimObs_from_sedgrid(
     # using both as the grid weight needed to account for the finite size
     #   of each grid bin
     # if we change to interpolating between grid points, need to rethink this
-    gridweights = sedgrid["weight"] * model_compl[:, filter_k]
+    gridweights = sedgrid[use_weight] * model_compl[:, filter_k]
     # need to sum to 1
     gridweights = gridweights / np.sum(gridweights)
 
--- a/beast/observationmodel/observations.py
+++ b/beast/observationmodel/observations.py
@@ -160,7 +160,7 @@ def gen_SimObs_from_sedgrid(
     compl_filter="F475W",
     ranseed=None,
     vega_fname=None,
-    use_weight='weight',
+    weight_to_use='weight',
 ):
     """
     Generate simulated observations using the physics and observation grids.
@@ -194,7 +194,7 @@ def gen_SimObs_from_sedgrid(
         filename for the vega info
         usefule for testing
 
-    use_weight : string (default='weight')
+    weight_to_use : string (default='weight')
         Set to either 'weight' (prior+grid), 'prior_weight', or 'grid_weight' to
         choose the weighting for SED selection.
 
@@ -229,7 +229,7 @@ def gen_SimObs_from_sedgrid(
     # using both as the grid weight needed to account for the finite size
     #   of each grid bin
     # if we change to interpolating between grid points, need to rethink this
-    gridweights = sedgrid[use_weight] * model_compl[:, filter_k]
+    gridweights = sedgrid[weight_to_use] * model_compl[:, filter_k]
     # need to sum to 1
     gridweights = gridweights / np.sum(gridweights)
 
--- a/beast/physicsmodel/prior_weights_stars.py
+++ b/beast/physicsmodel/prior_weights_stars.py
@@ -154,28 +154,37 @@ def compute_mass_prior_weights(masses, mass_prior_model):
       Unnormalized IMF integral for each input mass
       integration is done between each bin's boundaries
     """
-    # sort the initial mass along this isochrone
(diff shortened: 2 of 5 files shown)


beast/observationmodel/observations.py
```diff
@@ -160,6 +160,7 @@ def gen_SimObs_from_sedgrid(
     compl_filter="F475W",
     ranseed=None,
     vega_fname=None,
+    use_weight='weight',
```
karllark: Not a fan of the name of this keyword.  It sounds like usually the weight is not used unless you set this keyword.  Maybe `weight_to_use`?

docs/beast_priors.rst
```diff
@@ -136,6 +136,15 @@ The two mass function supported are:
...
 
+There is also a flat mass prior.  This is useful for creating grids for BEAST
+verification (see :doc:`Simulations <simulations>`), and should not be
+used for a standard fitting run.
+
+.. code-block:: python
+
+  mass_prior_model = {'name': 'flat_linear'}
+
+
```
karllark: Great.  The docs are here already.  Yeah.  I think you could just call this option `flat` as all the IMFs are evaluated in linear mass.  Saying `flat_linear` implies that linear is not always the case.

beast/physicsmodel/prior_weights_stars.py
```diff
@@ -154,29 +154,38 @@ def compute_mass_prior_weights(masses, mass_prior_model):
...
+
+        # calculate the average prior in each mass bin
+        for i in range(len(masses)):
+            mass_weights[sindxs[i]] = (quad(imf_func, mass_bounds[i], mass_bounds[i + 1]))[
+                0
+            ] / (mass_bounds[i + 1] - mass_bounds[i])
+
+    # if the prior is flat:
+    elif mass_prior_model["name"] == "flat_linear":
+        mass_weights = np.full(len(masses), 1.0)
```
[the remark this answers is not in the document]
lea-hagen: Happy to change it to a function.

beast/observationmodel/observations.py
lea-hagen: Putting this keyword in the category of things that make more sense in my brain than others' brains :).  `weight_to_use` sounds reasonable to me.

docs/beast_priors.rst
lea-hagen: Since the age priors distinguish between `flat_linear` and `flat_log`, I thought it would be good to be explicit.  But perhaps that's more of a thing for ages than masses?

lea-hagen:
Comments are incorporated, so should be good to go now.

docs/simulations.rst
```diff
@@ -26,19 +26,27 @@ of this file will determine the type of file output (e.g. filebase.fits for
...
+                --nsim 200 --compl_filter F475W
 
 The output file gives the simulated data in the observed data columns
 identified in the physicsgrid file along with all the model parameters
 from the physicsgrid file.  The names of the observed columns are
 `band_rate` and the units are normalized Vega fluxes (to match how
 the observed data are given).
 
+When creating simulated observations, using the standard IMF mass prior will
+skew your catalog to lower-mass stars.  If you wish to sample the full range of
```
karllark: Is the right word here "full"?  All the IMFs will sample the full range.  Maybe saying that if you want to have an equal number of stars in the simulation at all massed, then use the flat IMF is more accurate.

lea-hagen: I used "full" because I wasn't getting any stars above ~3 Msun.  But you're right, a flat IMF would be a better way to phrase it.

docs/simulations.rst
```diff
@@ -42,10 +42,10 @@ from the physicsgrid file.  The names of the observed columns are
 the observed data are given).
 
 When creating simulated observations, using the standard IMF mass prior will
-skew your catalog to lower-mass stars.  If you wish to sample the full range of
-stellar masses, you can set the mass prior to `{'name': 'flat'}` and the
-age prior to `{'name': 'flat_log'}` in `datamodel.py` before creating the
-model grid.
+skew your catalog to lower-mass stars.  If you wish to sample an equal number of
```
karllark: It is really an equal number?  I don't see how the two priors give an *equal* number.  I do see how this biases the sampling to give more high mass stars, but equal?

lea-hagen: Fair point.  I was paraphrasing your suggestion on the earlier iteration, but easy enough to change :)

karllark: Thanks.

PR #519, opened 2020-03-02
Updates to remove_filters.py

lea-hagen:
I've changed `tools/remove_filters.py` a bit so it's more flexible.  Now it can serve two use cases (I've copied the descriptions over from the docstring):
```
1. When making simulated observations, you want to test how your fit quality
   changes with different combinations of filters.  In that case, put in
   files for both `physgrid` and `obsgrid`.  Set `rm_filters` to the
   filter(s) you wish to remove, and they will be removed both from those
   and from the catalog file.  The three new files will be output with the
   name prefix set in `outbase`.

2. When running the BEAST, you have a master physics model grid with all
   filters present in the survey, but some fields don't have observations in
   all of those filters.  In that case, put the master grid in `physgrid`
   and set `rm_filters` to None.  The catalog will be used to determine the
   filters to remove (if any).  `obsgrid` should be left as None, because in
   this use case, the obsmodel grid has not yet been generated.  The output
   physics model grid will be named using the filename in `physgrid_outfile`
   (if given) or with the prefix in `outbase`.
```
I've updated the docs (both for Simulations and Workflow) with info about these options.

### DIFF
--- a/beast/tools/remove_filters.py
+++ b/beast/tools/remove_filters.py
@@ -12,54 +12,139 @@ from beast.physicsmodel.grid import FileSEDGrid, SpectralGrid
 import beast.observationmodel.noisemodel.generic_noisemodel as noisemodel
 
 
-def remove_filters_from_files(catfile, physgrid, obsgrid, outbase, rm_filters):
+def remove_filters_from_files(
+    catfile,
+    physgrid=None,
+    obsgrid=None,
+    outbase=None,
+    physgrid_outfile=None,
+    rm_filters=None,
+):
+    """
+    Remove filters from catalog, physics grid, and/or obsmodel grid.  This has
+    two primary use cases:
 
-    # remove the requested filters from the catalog file
+    1. When making simulated observations, you want to test how your fit quality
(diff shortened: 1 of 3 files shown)


beast/tools/remove_filters.py
```diff
@@ -12,70 +12,204 @@
...
-            cat.remove_column(colname)
+
+    # if rm_filters set, remove the requested filters from the catalog
+    if rm_filters is not None:
+        for cfilter in np.atleast_1d(rm_filters):
+            colname = "{}_rate".format(cfilter)
+            if colname.upper() in cat.colnames:
+                cat.remove_column(colname.upper())
+            elif colname.lower() in cat.colnames:
+                cat.remove_column(colname.lower())
```
karllark: I think you can replace these two if statements by one.  Wouldn't testing `colname.upper() in cat.colnames.upper()` cover both cases?

lea-hagen: The issue is that the name given to `remove_column` must be what's actually in the table.  I couldn't figure out a way around that other than the two ifs.

PR #526, opened 2020-04-02
Added docstring for pdf1d.py

s-goldman:
Added doc string for gen1d function


beast/fitting/pdf1d.py
```diff
@@ -86,6 +86,22 @@ def __init__(self, gridvals, nbins, logspacing=False, minval=None, maxval=None):
             self.pdf_bin_indxs = pdf_bin_indxs
 
     def gen1d(self, gindxs, weights):
+        """Determines the probabilities for each bin in weights.
```
karllark, line 89: What does it means "in weights"?  Do you mean "using the grid+prior weights and the likelihoods"?

karllark: And maybe "Calculate" instead of "Determine"?

karllark:
Thanks for this PR.  More docs are so welcome!

beast/fitting/pdf1d.py
s-goldman: I was trying to clarify that the "weights" here are the values for the combined (log likelihood, grid weights, and priors). Calculate works for me.

s-goldman: Also I thought the bins were determined by the grid weights. Hopefully that's correct and the correction makes that more clear.

beast/fitting/pdf1d.py
```diff
@@ -86,6 +86,22 @@ def __init__(self, gridvals, nbins, logspacing=False, minval=None, maxval=None):
             self.pdf_bin_indxs = pdf_bin_indxs
 
     def gen1d(self, gindxs, weights):
+        """Calculates the probabilities for each grid_weight bin.
```
karllark, line 89: What does "grid_weight" mean here?  The bins are for the posteriors PDFs so they are probabilities.  The weights are just a way to calculate the probabilities.  Basically a sum where the prior weight gives the prior probability, the grid weight gives the grid width (in some nD sense), and the log_likelihood gives the fitting probability.  Does that make sense?

I would say this sentence should be "Calculates the posterior probabilities for each bin."  Then the rest of the docstring says the details.

s-goldman: Yeah alright, I wan't sure how to word it. That sounds better. 

On a related note, is it true that for the bins, the parameter range that bin spans is set by the grid weights, or is that not true. I'm just trying to understand what it is doing.

karllark: Not quite sure I understand your question.  The range that the bins span is set by the grid.  The width of the bins is set by the grid bins, unless the number of bins needed to span all the unique grid points is larger than a set max.  In other words, if the number of unique grid points for a parameter is larger than the set max, then the max is used - otherwise the bins are set by the unique grid points for that parameter.  Does this make sense?  Did I actually answer your question?

s-goldman: I guess I was trying to better understand the bin widths. Are they just set by how the grid is spaced?

karllark: For those parameters that have less than the set max, yes.  For those that don't (e.g., mass), the bin widths are set by the max number of bins.

s-goldman: Got it, thanks for explaining it!

karllark:
Hummm...This PR was never merged.  @s-goldman : was there a reason you closed it?

Issue #531, opened 2020-04-17
info about tools that need tests


lea-hagen:
Lots of the tools need tests.  Since I've written/worked on many of them, here's a quick evaluation of what those need for testing.
* `compare_spec_type.py: could use the phat_small example with fake spectrally-typed stars+positions.  I made up a couple for the documentation that could be used.
* `cut_catalogs.py`: pretty straightforward, just need to compare to a cached version of the cut file.  Could potentially create tests with different combinations of input options (`partial_overlap`, `flagged`, etc).
* `read_beast_data.py`: contains 3 functions to test.  Each function reads in a file and outputs a dictionary (some only with requested keys), so it would probably be easy to make a test for this using the phat_small files.
* `remove_filters.py`: this has two different modes of removing filters (either for simulated files or for using a master grid), both of which need testing.  I think files would have to be created to run this on.
* `setup_batch_beast_fit.py` and `setup_batch_beast_trim.py`: These create input files for submitting to a queue.  Since they're just text files, they should be easy to test.
* `split_ast_input_file.py`: This file may not be necessary anymore (need to check with @benw1), so this is very low priority for testing.
* `star_type_probability.py`: adds up probability from 1D and 2D PDF files and outputs a dictionary or file.  Would need to run the BEAST to create 2D PDFs for phat_small and add those to the remote data, as well as adding the output file of probabilities to remote data for comparison.
* `write_sbatch_file.py`: just combines the input parameters into a text file, so should be very easy to test

lea-hagen:
I'll work on `compare_spec_type.py` testing today.

mdecleir:
I will look into write_sbatch_file.py for this last hour

lea-hagen:
And I just started on `star_type_probability.py`!

karllark:
I am working on `read_beast_data.py`

mdecleir:
I submitted PR #551 for the `write_sbatch_file.py` test.

PR #534, opened 2020-04-17
Fix for ambiguous filters in remove_filters

### LINKED ISSUE #527: remove_filters.py can't distinguish between instruments
Our catalog files only have a short filter name for the column names.  If a model grid has the same filter in two different instruments (e.g., F475W in both ACS and WFC3), neither of them get removed, because the filter name from the catalog file exists in both strings (e.g., `F475W` is in both `HST_ACS_WFC_F475W` and `HST_WFC3_F475W`).

This will also be a problem if the `rm_filters` keyword is set and a filter is in the grid with two instruments.

I know that at various points, we've talked about naming the catalog columns with the filter AND instrument, which would solve this.  Could also probably add an optional input to remove_filters.py that takes instrument names.

lea-hagen:
I've added an optional input to `remove_filters` to specify the BEAST filter associated with the short filter name.  Also, I simplified the code a bit so that instead of having some lists with info appended and some lists with info deleted, it now just does deleting.

I still need to confirm that it works and add docs (and tests!).

Closes #527

Partially addresses #531

### DIFF
--- a/beast/tools/remove_filters.py
+++ b/beast/tools/remove_filters.py
@@ -90,7 +90,7 @@ def remove_filters_from_files(
 
     # if rm_filters not set, extract the filter names that are present
     if rm_filters is None:
-        cat_filters = [f[:-5].upper() for f in cat.colnames if f[-4:].lower() == 'rate']
+        cat_filters = [f[:-5].upper() for f in cat.colnames if f[-4:].lower() == "rate"]
 
     # if beast_filt is set, make a list of the short versions
     if beast_filt is not None:
@@ -146,7 +146,6 @@ def remove_filters_from_files(
                             if cfilter in grid_col:
                                 rgridcols.append(grid_col)
 
-
             # --------------------------
             # if the removed filters are determined from the catalog file
             if rm_filters is None:
@@ -177,8 +176,6 @@ def remove_filters_from_files(
                         if cfilter in grid_col:
                             rgridcols.append(grid_col)
 
-
-
         # delete column(s)
         nseds = np.delete(g0.seds, rindxs, 1)
         nlamb = np.delete(g0.lamb, rindxs, 0)
@@ -197,7 +194,7 @@ def remove_filters_from_files(
         elif outbase is not None:
             g.writeHDF("{}_seds.grid.hd5".format(outbase))
         else:
-            raise ValueError('Need to set either outbase or physgrid_outfile')
+            raise ValueError("Need to set either outbase or physgrid_outfile")
 
     # if obsgrid set, process the observation model
     if obsgrid is not None:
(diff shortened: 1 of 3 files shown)


lea-hagen:
Two more changes:
* Added tests for both cases of removing filters
* To create those tests, I needed to extract the filter names from the SED grid, so I updated `read_beast_data.read_sed_data` to be able to return the list of filter names.

For the tests to run, I created a phat_small SED grid with an extra F475W column (so it has both ACS and WFC3).  @karllark, I'll send you the path to upload it.

lea-hagen:
I just resolved the merge conflicts, so this should be good to go (unless there are changes I should make, of course!).

beast/tools/remove_filters.py
```diff
@@ -93,41 +105,96 @@ def remove_filters_from_files(
...
+
+                        # if the current filter isn't in the BEAST ref list, keep it
+                        else:
+                            continue
+
+                    # if there isn't a list of BEAST refs, keep it
+                    else:
+                        continue
+
+                # if the current filter isn't in the catalog filters, delete it
```
karllark, line 185: Why are the continue statements needed?  It seems that the if statements could be reconfigured to provide the same results since everything(?) is already embedded in if statements.

beast/tools/remove_filters.py
```diff
@@ -93,41 +105,96 @@ def remove_filters_from_files(
...
+                                    if cfilter in grid_col:
+                                        rgridcols.append(grid_col)
+
+                        # if the current filter isn't in the BEAST ref list, keep it
+                        else:
+                            continue
+
+                    # if there isn't a list of BEAST refs, keep it
+                    else:
+                        continue
```
karllark, line 183: Are these statements even needed?  Isn't this the behavior of an `if` statement without an `else` statement?

lea-hagen:
@karllark Thanks for reviewing!  I agree that it's a bit over-complicated (I even had to draw out a flowchart to make sure I got it right), but given all of the ways the different options interact, I don't know if there's an obvious way to simplify it.  Perhaps splitting into two separate functions?

Regarding the continues: I put them in there to make it easier to follow the logic, though it's debatable whether that's useful for people other than me :).  Since the two use cases have opposite flows (when one deletes, the other doesn't), their two code chunks have a nice parallel structure that made editing more straightforward.  But now that it's written, if it would be more clear to delete the continues, I'm happy to do that!

karllark:
I would say it is more pythonic to not use continue statements unless they are explicitly needed.  And so would be in favor of removing them.

karllark:
You might wish to run `black` on your code.  It should improve the readability to get things more in a common format.

lea-hagen:
The continues are removed (hooray for tests to make sure the code still works as expected!), and I blackened the two files.

PR #539, opened 2020-04-17
Test for compare_spec_type.py

lea-hagen:
Testing `compare_spec_type` using the same spectrally-typed stars I invented for its documentation.  I'm just comparing the output dictionaries, so nothing too complicated.  Happy to do a more careful comparison of individual parameters if that would be useful.

This required adjusting the relative tolerance when comparing tables, so I also added a new input to `compare_tables` in beast/tests/helpers.py that lets you set `rtol`.

Addresses #531

### DIFF
--- a/beast/tests/helpers.py
+++ b/beast/tests/helpers.py
@@ -29,7 +29,7 @@ def download_rename(filename):
     return fname
 
 
-def compare_tables(table_cache, table_new):
+def compare_tables(table_cache, table_new, rtol=None):
     """
     Compare two tables using astropy tables routines.
 
@@ -38,6 +38,8 @@ def compare_tables(table_cache, table_new):
     table_cache : astropy table
     table_new : astropy table
         data for comparision.
+    rtol : float (default=None)
+
     """
     if not len(table_new) == len(table_cache):
         raise AssertionError()
@@ -46,11 +48,19 @@ def compare_tables(table_cache, table_new):
         # test numerical types for closeness
         # and other types for equality
         if table_new[tcolname].data.dtype.kind in ["f", "i"]:
-            np.testing.assert_allclose(
-                table_new[tcolname],
-                table_cache[tcolname],
-                err_msg=("%s columns not equal" % tcolname),
-            )
+            if rtol is None:
+                np.testing.assert_allclose(
+                    table_new[tcolname],
+                    table_cache[tcolname],
+                    err_msg=("%s columns not equal" % tcolname),
+                )
+            else:
+                np.testing.assert_allclose(
+                    table_new[tcolname],
+                    table_cache[tcolname],
+                    rtol=rtol,
+                    err_msg=("%s columns not equal" % tcolname),
+                )
         else:
             np.testing.assert_equal(
                 table_new[tcolname],
--- a/beast/tools/tests/test_compare_spec_type.py
+++ b/beast/tools/tests/test_compare_spec_type.py
@@ -1,0 +1,53 @@
+from beast.tools import compare_spec_type
+from beast.tests.helpers import download_rename, compare_tables
+from astropy.table import Table
+
+@remote_data
+def test_compare_spec_type():
+    """
+    Test for compare_spec_type.  The spectrally-typed stars aren't real sources,
+    they're just invented for the purposes of documenting/testing the code.
+    """
+
+    # download the needed files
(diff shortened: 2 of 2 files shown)


beast/tests/helpers.py
```diff
@@ -46,11 +48,19 @@ def compare_tables(table_cache, table_new):
...
+                    table_cache[tcolname],
+                    err_msg=("%s columns not equal" % tcolname),
+                )
+            else:
+                np.testing.assert_allclose(
+                    table_new[tcolname],
+                    table_cache[tcolname],
+                    rtol=rtol,
+                    err_msg=("%s columns not equal" % tcolname),
+                )
```
karllark, line 63: The if statement is not needed.  You can just use the statement in the `else` portion as if `rtol=None` then it is like not setting rtol at all.

beast/tools/tests/test_compare_spec_type.py
```diff
@@ -0,0 +1,106 @@
...
+    })
+
+    # compare to new table
+    compare_tables(expected_table, Table(spec_type))
+
+
+if __name__ == '__main__':
+
+    test_compare_spec_type_inFOV()
+    test_compare_spec_type_notFOV()
```
karllark, line 106: Not needed after the initial testing of the test code.  Best to remove if not needed to reduce future maintenance burden.

beast/tools/tests/test_compare_spec_type.py
```diff
@@ -0,0 +1,106 @@
+from beast.tools import compare_spec_type
```
karllark, line 1: Better to just import the specific function needed.  So this could be `from beast.tools.compare_spec_type import compare_spec_type` and then the function can be called with just `compare_spec_type` instead of `compare_spec_type.compare_spec_type`.

beast/tests/helpers.py
lea-hagen: I tried doing exactly that, but when `rtol=None`, `np.testing.assert_allclose` throws an error saying a float is required.

```
>>> np.testing.assert_allclose(1, 1.5, rtol=None)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-21697138e8e1> in <module>()
----> 1 np.testing.assert_allclose(1, 1.5, rtol=None)

~/miniconda3/envs/ac/lib/python3.5/site-packages/numpy/testing/_private/utils.py in assert_allclose(actual, desired, rtol, atol, equal_nan, err_msg, verbose)
   1489
   1490     actual, desired = np.asanyarray(actual), np.asanyarray(desired)
-> 1491     header = 'Not equal to tolerance rtol=%g, atol=%g' % (rtol, atol)
   1492     assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
   1493                          verbose=verbose, header=header, equal_nan=equal_nan)

TypeError: a float is required
```

karllark: That is not what I expected.  Weird.  Then I would suggest setting the default to 1e-7 as this is the same as the default rtol for this function.  Better than having the code duplicated I would say.

lea-hagen: sounds good!

beast/tools/tests/test_compare_spec_type.py
lea-hagen: I thought it was best coding practice to import the module rather than the function.  But I guess there's not really ambiguity in the name `compare_spec_type`!

beast/tests/helpers.py
```diff
@@ -38,6 +38,9 @@ def compare_tables(table_cache, table_new):
     table_cache : astropy table
     table_new : astropy table
         data for comparision.
+    rtol : float (default=1e-7)
+        relative tolerance for np.testing.assert_allclose (default of 1e-7
+        matches the default in assert_allclose)
     """
     if not len(table_new) == len(table_cache):
         raise AssertionError()
```
karllark, line 46: Shouldn't there be an error message with this error?  Usually done to provide a bit more info to the user.

karllark: I know this is not the code you wrote - I'm sure I did.  But could you add one?  Or is that not possible with an AssertionError?

lea-hagen: Looks like it's possible, so I'll go add a quick few words :)

PR #540, opened 2020-04-18
Docs, tests, and cleanup for beast_data

karllark:
Minor updates to beast_data as part of the MegaBEAST effort.
Tests and docs (via API) for beast_data.py.  Partially address #531.

Fixes "H5pyDeprecationWarning: dataset.value has been deprecated. Use dataset[()] instead." warning.

### DIFF
--- a/beast/tools/read_beast_data.py
+++ b/beast/tools/read_beast_data.py
@@ -8,6 +8,9 @@ import h5py
 from tqdm import tqdm
 
 
+__all__ = ["read_lnp_data", "read_noise_data", "read_sed_data", "get_lnp_grid_vals"]
+
+
 def read_lnp_data(filename, nstars=None, shift_lnp=True):
     """
     Read in the sparse lnp for all the stars in the hdf5 file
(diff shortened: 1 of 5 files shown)


karllark:
Could use tests.
Should check there are docs for this code.

karllark:
Should add beast_data to the API.

karllark:
@lea-hagen - great if you could do a review of this PR.  Note that the docs are via the API - see tools_api.rst.  This only renders once the docs are built, but it uses the docstring - automatic documentation.   Should do this for other tools (see #542), I might do this today even.

beast/tools/read_beast_data.py
```diff
@@ -132,39 +126,38 @@ def read_sed_data(
...
+    Two possible returns depending on param_list input
 
-    beast_data: dictonary
-       contains arrays of the requested SED grid parameters
+    sed_data: dictonary (param_list input as list of strings)
+        contains arrays of the requested SED grid parameters
+        if return_params is True, then also provides
+
+    grid_param_list : list of strings (param_list input as None)
+        if param_list is None, return the list of parameter options
```
lea-hagen, line 136: I think I forgot to update the docstring when I changed this in #513. If you could update this to say when `return_params is True` rather that `param_list` is `None`, that would be awesome.

karllark: Done.

karllark:
The key to getting the bold correct in the returns doc section is to have the space on both sides of the `:`.  Will have to fix this elsewhere in the code.  But at least I fixed it for this PR.

karllark:
@lea-hagen : ok to merge?

lea-hagen:
Awesome!  Yes, looks good to merge.

karllark:
Yeah!

PR #544, opened 2020-04-20
Test for star_type_probability

lea-hagen:
This adds a test for `star_type_probability.py`.  Funny enough, as part of doing this, I found a bug - when the required 2D PDF didn't exist, it returned None (instead of a list of None), so the results could be returned as a dictionary, but not saved as a table in a fits file.

This will require two files to be added to the remote data folder: `beast_example_phat_pdf2d.fits` and `beast_example_phat_startype.fits`.  @karllark, they add up to ~8 MB, so I'll email you their paths in my directory.

Partially addresses #531

### DIFF
--- a/beast/tools/star_type_probability.py
+++ b/beast/tools/star_type_probability.py
@@ -199,7 +199,8 @@ def ext_O_star(pdf2d_data, pdf2d_bins, min_M_ini=10, min_Av=0.5, max_Av=99):
         mass_bins = pdf2d_bins['M_ini+Av'][0,:,:]
     else:
         print("2D PDFs don't contain M_ini and Av data")
-        return None
+        tot_stars = pdf2d_data[pdf2d_data.keys()[0]].shape[0]
+        return [None] * tot_stars
 
     # reshape the arrays
     prob_data = prob_data.reshape(prob_data.shape[0], -1)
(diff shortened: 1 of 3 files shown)


karllark:
Yeah you found a bug.  One of the benefits of writing the tests is that when writing them you might find bugs!

karllark:
Is this ready for merging?  Or were you going to add additional tests?  I just looked at the coverage change and there is about a quarter of the tool file that is not covered.  Just checking.

lea-hagen:
Good question.  The other ~25% is from two things:
* Checking the order of parameters (e.g., `'M_ini+Av'` or `'Av+M_ini'`): This order is set by the order of parameters in the SED grid file, so I wouldn't expect it to change, but I suppose you never know.  It's also easier (and faster) than reading in a grid file to see what the order should be.  I'm not sure how to test this.
* Missing 2D PDF: I suppose I could do another phat_small run that doesn't save the needed parameters, and upload that as a separate cached file.

lea-hagen:
I just added another test for when the needed 2D PDFs aren't saved.  I'll email you the reference files again, @karllark.

I also had to change the returned value from None to NaN.  Updated the docs to include that.

PR #546, opened 2020-04-21
remove logL cut and irrelevant comments

galaxyumi:
logL = -9.999 are now dealt when generating the spec grid. Thus, there is no need to make logL cut when generating input ASTs. See #486 and #491

### DIFF
--- a/beast/observationmodel/ast/make_ast_input_list.py
+++ b/beast/observationmodel/ast/make_ast_input_list.py
@@ -133,15 +133,9 @@ def pick_models_toothpick_style(
     sedsMags = -2.5 * np.log10(modelsedgrid.seds[:] / vega_flux)
     Nf = sedsMags.shape[1]
 
-    idxs = np.where(modelsedgrid.grid["logL"] > -9)[0]
-    sedsMags_cut = sedsMags[idxs]
-
-    # Note that i speak of fluxes, but I've recently modified this to
-    # work with mags instead
-
     # Set up a number of flux bins for each filter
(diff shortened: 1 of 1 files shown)


karllark:
Would it be worth putting in a check for such cases here and then raising a warning/error if it happens?  Just wondering if this issue might happen again and silently sneak through.

galaxyumi:
> Would it be worth putting in a check for such cases here and then raising a warning/error if it happens? Just wondering if this issue might happen again and silently sneak through.

It might be safer. I added a warning raiser, and added back two lines to cut out logL=-9.999 SED models just in case.

beast/observationmodel/ast/make_ast_input_list.py
```diff
@@ -133,15 +133,16 @@ def pick_models_toothpick_style(
     sedsMags = -2.5 * np.log10(modelsedgrid.seds[:] / vega_flux)
     Nf = sedsMags.shape[1]
 
-    idxs = np.where(modelsedgrid.grid["logL"] > -9)[0]
-    sedsMags_cut = sedsMags[idxs]
-
-    # Note that i speak of fluxes, but I've recently modified this to
-    # work with mags instead
+    # Check if logL=-9.999 model points sliently sneak through
+    if min(modelsedgrid.grid["logL") < -9:
```
karllark, line 137: There is a missing `]` in this line right before the closing `)`

karllark:
Thanks for adding in this warning and the cut.  Great idea.  Syntax error needs fixing and then this will be ready for merging.

Issue #496, opened 2020-01-25
Supplement an existing AST list with an additional set


galaxyumi:
In some cases, there might be need to generate an additional set of AST inputs with specific conditions. For example, we may need more ASTs in a particular color-magnitude space (e.g., for running MATCH) or we may want to switch from an original coarse model grid to a denser model grid. To maximize the AST efficiency, we need to avoid sampling the same AST inputs that are already in the existing AST list.

PR #547, opened 2020-04-21
save SED grid index for selected ASTs

galaxyumi:
Save SED grid index for selected ASTs in the output "*_ASTparams.fits" file to track down which SED models were selected as input ASTs. This makes it possible to exclude these SEDs when supplementing an existing input ASTs. Related to Issue #496


lea-hagen:
Good idea to include that info!  Could you add a quick note about it in the docs?
https://beast.readthedocs.io/en/latest/generating_asts.html#returns

galaxyumi:
> Could you add a quick note about it in the docs?
> https://beast.readthedocs.io/en/latest/generating_asts.html#returns

Done!

PR #551, opened 2020-04-27
Test for the write_sbatch_file tool

mdecleir:
I wrote a test for the tool `write_sbatch_file` (see issue #531 ) It creates a new sbatch script, reads it in again, and checks if the read in string is the same as the expected string. However, I am really not sure if this is the best way to test this function, because the string is hardcoded, so if you change a little thing to the write_sbatch_file function, even a space or so, it will not work anymore. Also, I don't know how to split a very long string over multiple lines, so please improve this.


karllark:
Multi line strings are possible using """ quotes.
Check out: https://stackoverflow.com/questions/10660435/pythonic-way-to-create-a-long-multi-line-string

karllark:
You might want to the "2nd" answer in the stackoverflow.  s = (" ", " ", " ")

karllark:
Looks like a good test.  If the codes gets changed, then the test will need to be updated.  This is a standard part of the workflow - so no problems.

mdecleir:
I now used the """ method, which works, but looks a bit weird because of the lack of indentation. What do you think?

karllark:
Good enough.  We can always update later.  Thanks.  I'll ask Lea to review as she is the most knowledgeable about this code.

karllark:
@lea-hagen : if you have time to review this, great.  If you are busy, no worries, just tell me and I'll move forward with this PR.

lea-hagen:
I won't have time today, but I'll take a look tomorrow, if that works ok!

karllark:
Tomorrow would be great.

mdecleir:
> Looks great, @mdecleir! I just have one question/suggestion. I don't see the file `create_LMC_mastergrid.script` used anywhere, so can it be removed? Or, as another option, you could add it to @karllark's list of remote files, and then instead of the giant string for the expected output, you could just read in the file.

That file is indeed not really used, but each time one runs pytest, this file will be created and then read in. So, I am not sure what the best practice is here. I can remove it, but if someone runs pytest it will be created again.

lea-hagen:
Ok, I see!  Then it sounds like the file can be removed.  The idea with testing is that every time the tests get run, it creates that file, and checks if it has changed (by referencing the giant string).  So having a permanent version in the repository isn't needed.

karllark:
Yes, you can remove it.  It will be created during the tests.  You can make it a temporary file by using the python `tempfile` package.  Then the file will be automatically deleted when the testing code finishes.  I only recently learned about `tempfile` and am starting to use it in new test code I create.  Avoids having to specify `/tmp` as a path and nicely cleans up afterwards.

mdecleir:
I deleted the file now, but I am not sure how to implement the `tempfile` for the test, because the file is actually created in the tool `write_sbatch_file`. I guess we don't want this tool to always create a temporary file right?

karllark:
The `tempfile` returns the name of the file to be used.  So, you just pass that filename to the routine.

There is an example of how I've used this package.

```
tfile = NamedTemporaryFile(suffix=".hd5")
tgrid.writeHDF(tfile.name)
```

mdecleir:
> The `tempfile` returns the name of the file to be used. So, you just pass that filename to the routine.
> 
> There is an example of how I've used this package.
> 
> ```
> tfile = NamedTemporaryFile(suffix=".hd5")
> tgrid.writeHDF(tfile.name)
> ```

I understand now, thanks! I updated the test. It should now create a temporary file instead.

Issue #320, opened 2019-05-03
new datamodel class


lea-hagen:
I'm working on changing `datamodel.py` to a class.  It will take a text file as an input and create the `datamodel` class, and will be able to interpret python functions, strings, and units.  I'm also incorporating the verification code into it.

karllark:
Very cool.  This should really help in making things more understandable and usable (and robust!).

Issue #299, opened 2019-03-05
Major rework of BEAST files


karllark:
From the Feb 2019 BEAST HackDay, the idea emerged that a major rework of the BEAST file data formats could provide significant benefits.  One particular issue is the challenges of using the HDF5 format as it has hard for multiple BEAST developers to understand or manipulate.  The benefits could be easier to understand/manipulate, less code for reading/writing -> less code maintenance, and better suited for Mega-BEAST (and BEAST) needs.

Currently, the formats are a mix of ASCII csv, hdf5, and fits files.  

So, what would the format(s) of the BEAST files look like w/o the considering the current formats?  In other words, if we started with a clean slate, what would we do?  We understand the needs of the BEAST *and* Mega-BEAST much better now.

Please comment with ideas/proposals.

Existing issues that touch on this topic:
Output file format: #53 
Size of the files: #295
HDF5 issues; #186, #262 
unclosed files: #64 
eztables: #9, #10

karllark:
A file format that is more understandable than hdf5 would be nice.  Possibilities are FITS or ASDF.  But concerns about speed.

karllark:
Fewer files would be easier to manage.  Possibilities
1. Merge spec and spec_w_priors possible
2. Merge physics and observation models, but then harder to have separate observation (noisemodels) for a single physicsgrid file.
3. Merge output stats, pdf1d, and lnp files into a single file.  Ideally have one "record" for each star fit.

karllark:
For tables, we should use astropy.tables to take advantage of the extensive work done by the larger community.

lea-hagen:
I'd like to change the input file format: make `datamodel` a text file rather than `.py` (similar to what is done now for the megabeast).

karllark:
Good point.  Would be great to be able to just us a text file.  One impact would mean it would be easy to setup the beast to run from any directory with the BEAST scripts installed on the system.

karllark:
HDF5 info

python h5py closer to numpy than pytables:
http://docs.h5py.org/en/stable/faq.html#what-s-the-difference-between-h5py-and-pytables

pytables more database orientated and includes non-numpy data types:
http://www.pytables.org/FAQ.html#how-does-pytables-compare-with-the-h5py-project

karllark:
ASDF info

https://asdf-standard.readthedocs.io/en/stable/

karllark:
[...]

PR #553, opened 2020-04-30
Updates to grid classes and backends

### LINKED ISSUE #10: Table support from astropy instead of eztables
Astropy now includes a very nice interface to generic tables i/o and manipulation.  Current the BEAST uses a mix of astropy tables and eztables.  Better to just use one and astropy tables is supported by the community, thus no need to support it ourselves even as an external package.

karllark:
This large PR is the result of working to better understand the grids used in the BEAST.

The SpectralGrid and SEDGrid are now the two ways to access spectral and sed grids with the BEAST (aliases to these removed throughout the code).  The goal is to make it easier to know which one to use.  Developer documentation has been added to explain the two classes.  

Cleaning up docstrings and syntax (removing python 2 support).
One goal is to remove the dependence on eztables and pytables for the grids without breaking anything.  The classes now use astropy tables and h5py.
Another goal was to prepare the classes to support other disk file formats (e.g., asdf) transparently to the rest of the BEAST code.
This supports the megabeast as efficient access to the SED info will speed up its development and running.

A large amount of code has removed been as part of switching to using the community supported astropy table interface (instead of eztables).

Closes #10
Partially addresses #9
Supports #299, #320

The grid classes were originally written by @mfouesneau.  Kudos to him for setting things up such that we can abstract the grid access.  Not surprisingly given the time since these classes were defined, serious maintenance was needed to bring them up-to-date and make them ready future enhancements.


karllark:
Issue with tools/remove_filters.py.  Some memory communication between the 4 tests in the test_remov_filters.py with the shift away from eztables.  Currently do not understand.

karllark:
No idea why the c/c++ travis build is failing.  Not really important as that code was not touched in this PR.

mfouesneau:
What do you want me to do?

karllark:
@mfouesneau : nothing required.  Just wanted to tag you to acknowledge all the work you did in this area in the past.  If you want to look at this PR, you are welcome of course.  But this is not needed and I know you are busy.

karllark:
@lea-hagen and @galaxyumi.  You both are two that have been working with the BEAST recently and are the obvious people that might have time and motivation to review this PR.  This is a big PR, so I know this is a big ask.  There are tests, so I'm pretty sure this should not break anything.

lea-hagen:
Wow, this is a huge effort!  Thanks @karllark!  I should have time to review it next week.

I recall that eztables had a lot of optimizations to improve read/write speed.  Do you have a sense about whether run times have changed significantly by switching to astropy tables?

karllark:
I think h5py has similar optimizations.  But I'm not sure.  Investigating this can be a separate image.  Such optimizations also exist for FITS files via memmapping.  I haven't fully implemented these (e.g., for the disk backend).  But I don't think we are taking advantage of much of any of these optimizations at present.  Most of our access to the grids is via the memory backend.  Using the cache and disk backends would be needed to use such I think.

lea-hagen:
I started reviewing and come across an early problem.  With moving to astropy tables, there are commands in datamodel that don't work anymore.  Specifically, in the `GenFluxCatalog` class, there's a line in datamodel that sets aliases:
```
        for ik,k in enumerate(filters):
            self.data.set_alias(k, obs_colnames[ik])
```
This class is replicated in `beast/fitting/tests/helpers.py`, where those lines have been updated to:
```
        for ik, k in enumerate(filters):
            self.filter_aliases[k] = obs_colnames[ik]
```
So in the short term, we'll all need to update our datamodel files to incorporate the changes to this class.  Longer term, this class should of course be moved out of datamodel, as we've discussed many times.  Given that it already exists as a helper (where it's imported for multiple tests), could it be moved somewhere else and used instead of looking for it in datamodel?  Though that might be beyond the scope of this PR.

karllark:
Yeah, this was an issue I found.  The `set_alias` member function being used is part of eztables.  For this PR, I was trying to avoid changing more than needed.  Since there is a solution that "just" involves updating the datamodel.py, I decided that was the way to go.  This does remind me that I have a PR I need to submit for beast-examples that does this fix.

I agree a better solution is needed and would make things easier long-term.  I find the filter aliases confusing and would like a better solution.  One of the challenging things is that we need the translation between the internal filter names (e.g., HST_ACS_F475W), the observed names (e.g., F475W), and the asts filters names (e.g., f475w in/out versions).  I think a better solution is to provide this information in the datamodel.py and have the `Observation` and `AST` classes (latter doesn't exist I think) take that info and use it.  Then we could get rid of the defining the aliases in functions/classes in datamodel.py.  But as you say, this should probably be a separate PR.

Note, this PR will not be part of v1.4.  Going to release what we have as I think this PR starts breaking compatibility and so is the really the start of BEAST v2.0.

beast/physicsmodel/grid.py
```diff
@@ -217,28 +207,28 @@ def getSEDs(
...
-        inplace:bool
+        inplace : bool, optional
             if set, do not copy the grid and apply extinction on it
 
-        filterLib:  str
+        filterLib : str, optional
             full filename to the filter library hd5 file
 
-        **kwargs extra keywords will be forwrded to extLaw
+        **kwargs extra keywords will be forworded to extLaw
```
galaxyumi, line 243: please fix a typo: 'forwarded'

karllark:
@galaxyumi - many thanks for the review!  I will fix the typo.

lea-hagen:
I can't make a comment directly in the file (because it's too far from diff-ed lines), but in Line 78 of beast/tools/trim_many_via_obsdata.py, `datamodel.get_obscat` needs an additional input.

beast/fitting/tests/helpers.py
```diff
@@ -0,0 +1,100 @@
...
+    """ Function that generates a data catalog object with the correct
+    arguments
+
+    Parameters
+    ----------
+    obsfile: str, optional (default datamodel.obsfile)
+        observation file
+
+    filters: sequence(str), optional, datamodel.filters
+        seaquence of filters of the data
```
lea-hagen, line 92: Could you add info about the other input parameters?

karllark:
> In addition to the small changes I already mentioned above, I found that `run_beast.py` in phat_small will need updating - it uses the old `FileSEDGrid`.

Good point.  Checkout the newly created https://github.com/BEAST-Fitting/beast-examples/pull/10

lea-hagen:
I think I figured out the bug.  But first, I would like to clarify if I'm understanding a change correctly.  In v1.4, `FileSEDGrid` (which just passed its inputs into `SpectralGrid`) was used for both the grid of spectra and the grid of band-integrated SEDs.  But now that doesn't exist anymore, and instead, there are two separate classes (`SpectralGrid` for spectra and `SEDGrid` for SEDs).  Is this correct?  If this is indeed the case, it makes it a a bit complicated to replace instances of `FileSEDGrid` with the right class (by which I mean we can't just use find-replace).

karllark:
> I think I figured out the bug. But first, I would like to clarify if I'm understanding a change correctly. In v1.4, `FileSEDGrid` (which just passed its inputs into `SpectralGrid`) was used for both the grid of spectra and the grid of band-integrated SEDs. But now that doesn't exist anymore, and instead, there are two separate classes (`SpectralGrid` for spectra and `SEDGrid` for SEDs). Is this correct? If this is indeed the case, it makes it a a bit complicated to replace instances of `FileSEDGrid` with the right class (by which I mean we can't just use find-replace).

That is correct.  Both work very similarly.  `SpectralGrid` just has a couple of extra member functions. I thought it would be good to make the distinction so that future development that would make them different could be supported.

beast/tools/run/create_physicsmodel.py
```diff
@@ -155,7 +156,7 @@ def create_physicsmodel(nsubs=1, nprocs=1, subset=[None, None]):
 
         # function to process the subgrids individually
         def gen_subgrid(i, sub_name):
-            sub_g_pspec = FileSEDGrid(sub_name)
+            sub_g_pspec = SEDGrid(sub_name)
```
lea-hagen, line 159: This should be `SpectralGrid(sub_name)` instead.

lea-hagen:
I managed to get through the entire phat_small example with subgrids.  I found a small bug in `create_obsmodel`, which I'll make a separate PR for (since it should be in v1.4 too), but aside from that, once the `SpectralGrid` line is changed, this PR looks good to me!

beast/physicsmodel/grid.py
karllark: Done.

karllark:
> I can't make a comment directly in the file (because it's too far from diff-ed lines), but in Line 78 of beast/tools/trim_many_via_obsdata.py, `datamodel.get_obscat` needs an additional input.
> 
> Edit: looks like this also needs to be updated for the calls to `get_obscat` in `make_ast_inputs.py` and `run_fitting.py` in tools/run/.

What is the extra input?  The datamodel code for the phat-small example is giving what is needed.  Code on line 78 is: `obsdata = datamodel.get_obscat(obsfile, modelsedgrid.filters)`

beast/tools/run/create_physicsmodel.py
karllark: Done.

lea-hagen:
> What is the extra input? The datamodel code for the phat-small example is giving what is needed. Code on line 78 is: `obsdata = datamodel.get_obscat(obsfile, modelsedgrid.filters)`

Sorry, I should have pasted that in!  Should be
```obsdata = datamodel.get_obscat(obsfile, modelsedgrid.filters, datamodel.obs_colnames)```

When you say the example is running, do you mean the tests?  If so, the test does have the third input (which is how I figured out how to fix it) (example [here](https://github.com/BEAST-Fitting/beast/blob/2081b8ad3b69ae61a90a622182d07af8eedea426/beast/fitting/tests/test_fit_grid.py#L40))

File: beast/fitting/tests/helpers.py (hunk 2 above)
Comment (karllark) [reply beast/fitting/tests/helpers.py:92]: Done.

**Comment by karllark** [comment]
> Sorry, I should have pasted that in! Should be
> `obsdata = datamodel.get_obscat([existing_input], [existing_input], datamodel.obs_colnames)`
> 
> When you say the example is running, do you mean the tests? If so, the test does have the third input (which is how I figured out how to fix it) (example [here](https://github.com/BEAST-Fitting/beast/blob/2081b8ad3b69ae61a90a622182d07af8eedea426/beast/fitting/tests/test_fit_grid.py#L40))

Ahh...now this makes sense.  The tests and phat_small example runs fine.  But that is because the tests and phat_small.datamodel define things appropriately. 

I'm now thinking of updating and moving the GenFluxCatalog and get_obscat code out of the tests subdir and making it a generic part of the observations.  This would mean we can remove it from any and all datamodels.  That would be great and get us closer to having a datamodel that is a text file and not code.  Going to try and do this now as this code is already moved around in this PR.

**Comment by lea-hagen** [comment]
Moving `GenFluxCatalog` and `get_obscat` would certainly help things.  I'm trying to figure out if there's a way to incorporate that into v1.4, but I'm unconvinced.

**Comment by karllark** [comment]
> Moving `GenFluxCatalog` and `get_obscat` would certainly help things. I'm trying to figure out if there's a way to incorporate that into v1.4, but I'm unconvinced.

Maybe.   But probably not worth it at this point.  Things work.  We can make BEAST v2.0 really better and cleaner and ...

**Comment by karllark** [comment]
@galaxyumi : done with your review of this PR?

@lea-hagen : I made the update of not just moving `GenFluxCatalog`, but removing it.  What do you think?  It made more sense to just move the bit of what was in that class into the `Observations` class.  We only use the catalogs one way and having a derived class from `Observations` was just not needed.  And I don't know what I was thinking when I wrote `get_obscat` - it is just the same as using `Observations` or `GenFluxCatalog`.  Just goes to show my python learning curve.  So, now less code - easier to maintain.  And no functions/classes in the datamodel.py!

**Comment by lea-hagen** [comment]
Switching to `Observations` seems reasonable to me!  I'll chug through my source density and subgrid examples again to make sure those still work as expected.

File: beast/fitting/tests/test_fit_grid.py [hunk 4]
```diff
@@ -136,7 +38,7 @@ def test_fit_grid():
     basefilters = ["F275W", "F336W", "F475W", "F814W", "F110W", "F160W"]
     obs_colnames = [f.lower() + "_rate" for f in basefilters]
 
-    obsdata = get_obscat(obs_fname, filters, obs_colnames, vega_fname=vega_fname)
+    obsdata = Observations(obs_fname, filters, obs_colnames, vega_fname=vega_fname)
     # output files
     stats_fname = "/tmp/beast_example_phat_stats.fits"
     pdf1d_fname = "/tmp/beast_example_phat_pdf1d.fits"
```
galaxyumi, line 44: Do we want to add pdf2d_fname as well?

karllark: We should add checking the pdf2d results as well.  But that could be done in a separate PR.  Probably better that way as long as an issue is created so we don't forgot.  Does this sound reasonable?

beast/physicsmodel/grid.py
```diff
@@ -186,22 +152,62 @@ def __getattr__(self, name):
...
 
     def copy(self):
         """ returns a copy of the object """
         return self.__class__(backend=self._backend.copy())
 
 
+class SEDGrid(ModelGrid):
+    """
+    Generate a grid that the full observational model (SEDs).
+    Currently a directy interface to ModelGrid.  Setup for later expansion.
```
galaxyumi, line 165: typo: directly

beast/physicsmodel/grid.py
```diff
@@ -186,22 +152,62 @@ def __getattr__(self, name):
...
+    seds : ndarray
+        2D `float` array (# models, # bands) giving the seds
+
+    lamb : ndarray
+        1D `float` array of the wavelengths of the sed bands
+
+    filters : list
+        list of the filter names of the sed bands
+
+    grid : astropy.Table
```
galaxyumi, line 178: satrapy.table

beast/observationmodel/ast/make_ast_input_list.py
```diff
@@ -313,7 +313,7 @@ def pick_models(
         vega_f, vega_flux, lamb = v.getFlux(filters)
 
     # gridf = h5py.File(sedgrid_fname)
```
galaxyumi, line 315: We might want to remove the 3 lines including 'gridf' in this code that are currently all commented. I think we don't use 'gridf' anymore.

beast/observationmodel/ast/make_ast_xy_list.py
```diff
@@ -48,9 +48,8 @@ def pick_positions_from_map(
     Parameters
     ----------
 
-    catalog: GenFluxCatalog object
-        The output object from `datamodel.get_obscat(datamodel.obsfile,
-        datamodel.filters)`
+    catalog: Observations object
+        Provides the observations
 
     chosen_seds: astropy Table
```
galaxyumi, line 54: astropy Table -> astropy.table?

beast/observationmodel/ast/make_ast_xy_list.py
```diff
@@ -336,43 +366,54 @@ def pick_positions_from_map(
                     if catalog_boundary_radec:
                         within_bounds = catalog_boundary_radec.contains_points(
                             [[x, y]]
-                        )[0]  # N,2 array of AST X and Y positions
+                        )[
+                            0
```
galaxyumi, line 370: Is this line change necessary?

beast/observationmodel/ast/make_ast_xy_list.py
```diff
@@ -336,43 +366,54 @@ def pick_positions_from_map(
...
+                    [x], [y] = wcs.all_world2pix(
+                        np.array([ra]), np.array([dec]), wcs_origin
+                    )
 
                     # check that this x/y is within the catalog footprint
-                    within_bounds = catalog_boundary_xy.contains_points(
-                        [[x, y]]
-                    )[0]  # N,2 array of AST X and Y positions
+                    within_bounds = catalog_boundary_xy.contains_points([[x, y]])[
+                        0
```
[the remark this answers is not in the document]
lea-hagen: I agree.  I'm assuming it's just a strange outcome from blackening the code?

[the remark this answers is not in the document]
karllark: Yep.  I'll investigate and turn off the formatting for this line if I can't solve it some other way.

beast/observationmodel/noisemodel/toothpick.py
```diff
@@ -176,7 +188,7 @@ def _compute_sigma_bins(
         good_bins = np.zeros(nbins, dtype=int)
 
         # get the indexs to the recovered fluxes
-        good_indxs, = np.where(flux_out != 0.0)
+        (good_indxs,) = np.where(flux_out != 0.0)
```
galaxyumi, line 191: Do we want to use math.isclose for comparing floats?

beast/physicsmodel/creategrid.py
```diff
@@ -299,7 +300,7 @@ def make_extinguished_grid(
     if isinstance(spec_grid, str):
         ext = spec_grid.split(".")[-1]
         if ext in ["hdf", "hd5", "hdf5"]:
-            g0 = SpectralGrid(spec_grid, backend="hdf")
+            g0 = SpectralGrid(spec_grid, backend="disk")
```
galaxyumi, line 303: It's unclear to me which backend option needs to be used in which case.

beast/observationmodel/noisemodel/toothpick.py
lea-hagen: In this case, a flux of precisely zero is what indicates a source wasn't recovered.  I don't know if there's ever a chance of floating point errors messing that up, but I would suspect not, since `flux_out` is read directly from the fits file.

mfouesneau: Why the parenthesis on the left-hand side? my editor says this is not necessary. Is this a python 2.x thing? This may be better to use np.where(...)[0]??

karllark: This is inserted by `black` when I run it.  Been using this to auto-format code to avoid having to remember all the bloody PEP8 stuff.  It is just an explicit way to say it is a tuple.  There is differences of opinion in the community.  `black` picks a solution and avoids us having to decide.

beast/physicsmodel/creategrid.py
karllark: Check out the new docs in `docs/development_classes.rst`.  Hopefully this answers your question.

beast/physicsmodel/helpers/gridbackends.py
```diff
@@ -1,72 +1,85 @@
...
-HDFBackend:
-    Works directly with an HDFStore support, ie., on disk. Cache and reading
-    are allowed through any way offered by pytables, which becomes very handy
+DiskBackend:
+    Works directly with an h5py support, ie., on disk. Cache and reading
+    are allowed through any way offered by h5py, which becomes very handy
     for very low-memory tasks such as doing single star figures.
+"""
+import sys
+import astropy.io.fits as fits
```
galaxyumi, line 26: import astropy.io.fits as fits -> from astropy.io import fits

beast/physicsmodel/helpers/gridbackends.py
```diff
@@ -1,72 +1,85 @@
...
+import h5py
+import copy
+from astropy.table import Table
+from astropy.io.misc.hdf5 import read_table_hdf5
```
galaxyumi, line 30: Do we explicitly need hdpy and read_table_hdf5 in addition to astropy.table to read/write a HDF5 file?

karllark: Yes.  The read_table_hdf5 allows reading a table from a hdu in a hdulit.  E.g., in memory w/o opening the file a 2nd time.  I couldn't figure out how to do this with the standard Table.read command.  There might be a way, but...

beast/physicsmodel/stars/isochrone.py
```diff
@@ -342,7 +368,8 @@ def interpolation(self, b=None):
             return self.interp
 
     def _load_table_(self, source):
-        self.data = Table(self.source).selectWhere("*", "isfinite(logA)")
+        tdata = Table.read(self.source, format="csv", header_start=_get_num_comments(self.source))
```
galaxyumi, line 371: Can we use 'comment' parameter here instead of using 'get_num_comments' function?

karllark: Can you be more explicit in this suggestion?  I couldn't figure out how to tell the Table.read function where to get the header column names for the specific format of the saved isochrone files.  So I created this the `_get_num_comments` function to determine the number of comment lines in the file.  I tried other ways, but only got this to work.  Happy to do this a different way if there is one.

beast/plotting/plot_ast_histogram.py
```diff
@@ -30,83 +30,85 @@ def plot_ast(ast_file, sed_grid_file=None):
...
     # if chosen, read in model grid
     if sed_grid_file is not None:
-        modelsedgrid = FileSEDGrid(sed_grid_file)
+        modelsedgrid = SEDGrid(sed_grid_file)
         with Vega() as v:
             _, vega_flux, _ = v.getFlux(filter_cols)
         sedsMags = -2.5 * np.log10(modelsedgrid.seds[:] / vega_flux)
-        good_seds = np.where(modelsedgrid.grid['logL'] > -9)[0]
-
+        good_seds = np.where(modelsedgrid.grid["logL"] > -9)[0]
```
galaxyumi, line 49: We don't need this when we apply this cut while generating a SpectralGrid.

beast/tools/remove_filters.py
```diff
@@ -181,18 +181,18 @@ def remove_filters_from_files(
...
-            g0.grid.delCol(rcol)
+            g0.grid.remove_column(rcol)
 
         print("orig filters: {}".format(" ".join(filters)))
         print(" new filters: {}".format(" ".join(nfilters)))
 
         # save the modified grid
-        g = SpectralGrid(np.array(nlamb), seds=nseds, grid=g0.grid, backend="memory")
-        g.grid.header["filters"] = " ".join(nfilters)
+        g = SEDGrid(np.array(nlamb), seds=nseds, grid=g0.grid, backend="memory")
```
galaxyumi, line 190: We might want to keep using the 'SpectralGrid' here?

karllark: Can you say more about why we would use `SpectralGrid` instead?  My thinking was that if we are removing filters, then this is an `SEDGrid` as it isn't about spectra at this point.

beast/plotting/plot_ast_histogram.py
karllark: Good point.

beast/fitting/tests/test_fit_grid.py
karllark: I've created issue #566.

beast/physicsmodel/grid.py
karllark: Done.

beast/observationmodel/ast/make_ast_xy_list.py
karllark: Leaving it as is.  It really is astropy.table.Table, but that looks odd.  There is a way to reference the actual api for this class, but I haven't figured that out yet.  If I do, I'll use that method.  Googling now.

karllark: Ok.  That looks hard.  Running away.

beast/observationmodel/ast/make_ast_xy_list.py
karllark: Nope.  Formatting turned off for this line.

beast/observationmodel/ast/make_ast_xy_list.py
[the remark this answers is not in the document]
karllark: This auto-formatting happens for a number of lines due to the long variable names and how the tuple results are unpacked.  Instead of adding comment lines for each case to turn the formatting on/off, I just made the destination variable name `inbounds` instead of `within_bounds`.  This solves the problem and the variable name still means the same thing (I would posit).

beast/observationmodel/noisemodel/toothpick.py
karllark: For more info: https://github.com/psf/black/issues/1108

beast/physicsmodel/helpers/gridbackends.py
karllark: Good point.  Done.

beast/plotting/plot_ast_histogram.py
karllark: Removed this line and the use of `good_seds` in one place.

beast/physicsmodel/grid.py
karllark: I did figure out how to reference the astropy.table.Table class with a link!  So now clicking takes one to the Table class API.  Yeah.

beast/tools/remove_filters.py
galaxyumi: I think you are right. My brain used an automated mapping between the old and new codes: 'FileSEDGrid' to 'SEDGrid' and 'SpectralGrid' to 'SpectralGrid'.

beast/physicsmodel/stars/isochrone.py
galaxyumi: For an isochrone file, 
tdata = Table.read('beast_example_phat_iso.csv', format="csv", comment='#') 
gives me the tdata.colnames as follows:
['logL', 'logg', 'mbolmag', 'logA', 'logT', 'M_ini', 'M_act', 'stage', 'Z']
Is this doing what you want?

beast/tools/remove_filters.py
karllark: Ok.  Thanks for answering.

beast/physicsmodel/stars/isochrone.py
karllark: Awesome!  Most excellent.  Thanks.  I've updated the code, removed my `_get_num_comments` hack and am re-running all the tests.  Yeah!

beast/physicsmodel/grid.py
```diff
@@ -70,10 +70,11 @@ def __init__(self, *args, **kwargs):
...
             if provided, update the grid table aliases
 
-        backend : str or GridBackend(or subclass), optional
-            corresponding backend class
-            'memory': MemoryBackend,
-            'cache': CacheBackend,
+        backend : str or :class:`~beast.physicsmodel.helpers.gridbackends.GridBackend` subclass, optional
+            if str corresponding backend class
+            'memory' = MemoryBackend,
+            'cache' - CacheBackend,
```
lea-hagen, line 76: Not a big deal at all, but should this be `=` instead of `-`?

beast/physicsmodel/grid.py
```diff
@@ -162,7 +163,7 @@ def copy(self):
 class SEDGrid(ModelGrid):
     """
     Generate a grid that the full observational model (SEDs).
-    Currently a directy interface to ModelGrid.  Setup for later expansion.
+    Currently a directly interface to ModelGrid.  Setup for later expansion.
```
lea-hagen, line 166: direct?

Issue #554, opened 2020-05-04
Testing code in beast/tools/run


lea-hagen:
I'm starting to look at how to do testing for subgrids and/or source density splitting.  As a first step, I want to check the SED grids, which are created in `create_physicsmodel.py`.  But it looks like tests are currently written to test each step of the code independently (the spectral grid, prior/grid weights, final extinguished grid).  I could copy-paste the relevant parts of `create_physicsmodel` into a testing script, but the way it's written, it would require a fair amount of restructuring, to the point where I'd be concerned that the test isn't representative of what's actually happening in the original code.  Would it be appropriate to just have a test that runs `create_physicsmodel`, and then do a bunch of assert statements afterwards for each of the intermediate file products?

karllark:
Not sure.  This sounds like it will do many of the same tests as are already done.  Unless there is a way to just test the parts that are not provided by the current tests, maybe this code does not need tests?  At least at this point.

There is a limit to the execution time for tests on travis and I worry we will hit the limit.

lea-hagen:
> Not sure. This sounds like it will do many of the same tests as are already done. Unless there is a way to just test the parts that are not provided by the current tests, maybe this code does not need tests? At least at this point.

Hmm, good point.  I think it might work to download the files for the earlier steps, so it'll skip generating those, and just run the part where it makes the extinguished grid.  Since this is the only code that's set up to make subgrids, I want to at least test that part.

> There is a limit to the execution time for tests on travis and I worry we will hit the limit.

What's the limit?  My plan was to do the same grid settings as phat_small, and make it into two subgrids, so I wouldn't expect it to be too much longer.

lea-hagen:
I found the time limit info [here](https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts):
> Travis CI has specific time limits for each job, and will stop the build and add an error message to the build log in the following situations:
> * When a job produces no log output for 10 minutes.
> * When a job on a public repository takes longer than 50 minutes.
> * When a job on a private repository takes longer than 120 minutes.

Given the 50 minute limit per job, I think running the whole `create_physicsmodel` script will be fine.  Can probably create individual jobs for each of the other scripts in tools/run/ also.

karllark:
Go for it.  We can always disable tests that take too long.  :-)

lea-hagen:
[...]

PR #558, opened 2020-05-08
spectral grid test - remove isochrones with logL < -9

lea-hagen:
As discussed in #554, the tests didn't remove isochrones with `logL < -9` (#486, #491, #500), so this ensures the tests match what we expect.

### DIFF
--- a/beast/physicsmodel/stars/tests/test_spectral_grid.py
+++ b/beast/physicsmodel/stars/tests/test_spectral_grid.py
@@ -28,6 +28,9 @@ def test_make_kurucz_tlusty_spectral_grid():
     # read in the cached isochrones
     oiso = ezIsoch(iso_fname)
 
+    # remove the isochrone points with logL=-9.999
+    oiso = ezIsoch(oiso.selectWhere("*", "logL > -9"))
+
     # define the distance
     distances = [24.47]
     distance_unit = units.mag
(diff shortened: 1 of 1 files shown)


karllark:
Updating the cached files for this PR will break the tests for PR #553.  Ok to wait till that PR is merged to update the cached files?

beast/physicsmodel/stars/tests/test_spectral_grid.py
```diff
@@ -28,6 +28,9 @@ def test_make_kurucz_tlusty_spectral_grid():
     # read in the cached isochrones
     oiso = ezIsoch(iso_fname)
 
+    # remove the isochrone points with logL=-9.999
+    oiso = ezIsoch(oiso.selectWhere("*", "logL > -9"))
+
```
karllark, line 33: Note that this way of filtering the isochrones will not work once PR #553 is merged.  This relies on eztable specific functions.  New method (once #553 is merged) is: `oiso.data = oiso[oiso["logL"] > -9]`.

karllark:
Or I could move this change into the #553 PR.

lea-hagen:
Either way is fine!

lea-hagen:
I updated the syntax, but if it's easier to put this change directly into #553, that's totally fine.

karllark:
Great.  Will update the cache files after #553 is merged.

lea-hagen:
It looks like this can be merged now.  @karllark, I'll regenerate the reference files to make sure they're compatible with the changes in #553, and then email you with their paths.

lea-hagen:
Note to future self - files can be regenerated once #576 is resolved.

PR #559, opened 2020-05-11
Test for create_physicsmodel (tests for subgrids)

lea-hagen:
This adds two tests for `create_physicsmodel.py`: creating the physicsmodel grids with and without subgrids.  It uses the phat_small grid parameters for both, and the subgrid version uses two subgrids.  Once @karllark uploads the new remote data files, I may need to update the paths to the cached files.  Also, because `create_physicsmodel.py` removes stars with `logL < -9` (#554) and the existing cached files shouldn't be updated till #553 is merged (https://github.com/BEAST-Fitting/beast/pull/558#issuecomment-626201395), the tests will initially fail.

I also updated `compare_hdf5` in the beast/tests/helpers.py because the dataset syntax issue was making tests fail for me, but I just noticed that's also been changed in #553.  Hopefully that won't confuse/break anything!


karllark:
@lea-hagen : I am doing a somewhat large revision to the regression tests right now.  This is to take advantage of the library files being downloaded for the settings work.  I download them in other files directly.  Now that they are downloaded using the get_libfiles, we don't have to do this anywhere else.  But this requires moving all the regression tests to a single file and putting them in a class.

This is a warning that tests and files will be moving around.  This may (will?) cause problems for this PR.  Apologies in advance.  Happy to talk about this if that would help.

PR #571, opened 2020-06-01
Increase number of PDF bins

### LINKED ISSUE #570: Change default `max_nbins` for PDFs?
I just (re-)noticed that there's a parameter in `Q_all_memory` (beast/fitting/fit.py), `max_nbins`, that sets the maximum number of bins for the 1D and 2D PDFs.  The current default is 50.  So even if the grid is more finely sampled, the PDFs will have at most 50 bins.  Does this default still make sense, or should it be increased for production runs?  I'm unsure if the coarser PDF resolution will impact science goals.  Currently, their only use is for doing cuts for star types.

lea-hagen:
Per #570, increasing the max number of bins from 50 to 100.  Happy to change this to 200 (or some other number) if preferred.  This change should be added to both v1.4 and v2.0.

I expect that this will change the stats (slightly) and PDFs for non-gridded parameters (mass, luminosity, temperature, etc), so the tests will probably fail.  We'll have to coordinate updating files with #558 and #559.

Closes #570.

### DIFF
--- a/beast/fitting/fit.py
+++ b/beast/fitting/fit.py
@@ -338,7 +338,7 @@ def Q_all_memory(
     gridbackend : str or grid.GridBackend
         backend to use to load the grid if necessary (memory, cache, hdf)
         (see beast.core.grid)
-    max_nbins : int
+    max_nbins : int (default=100)
         maxiumum number of bins to use for the 1D likelihood calculations
     save_every_npts : int
         set to save the files below (if set) every n stars
@@ -885,6 +885,7 @@ def summary_table_memory(
     save_every_npts=None,
     lnp_npts=None,
     resume=False,
+    max_nbins=100,
     stats_outname=None,
     pdf1d_outname=None,
     pdf2d_outname=None,
@@ -921,6 +922,8 @@ def summary_table_memory(
     use_full_cov_matrix : bool
         set to use the full covariance matrix if it is present in the
         noise model file
+    max_nbins : int (default=100)
+        maxiumum number of bins to use for the 1D likelihood calculations
     stats_outname : str
         set to output the stats file into a FITS file with extensions
     pdf1d_outname : str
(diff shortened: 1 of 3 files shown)


lea-hagen:
@karllark it looks like a setting needs to be changed in Codacy to let it analyze the v1.4 branch.

karllark:
Codacy for v1.4 enabled.

karllark:
One of the tests failed - `test_fit_grid.py`.  Can you set the max_nbins in the tests to 50?  Then the tests will continue to work and be robust to future changes to max_nbins.

lea-hagen:
That sounds good.  Since `max_nbins` only exists in `Q_all_memory`, I'll need to propagate it through all the layers that call it, but that should be straightforward.  The question, though, is whether to add it as an input to `tools/run/run_fitting.py` or put it in datamodel.

karllark:
It should at least be propagated so that the tests can be run.  I'm not sure that this needs to be exposed in datamodel.py - at least at this point.  Not sure we want everyone to have to decide what value they want for this parameter.

lea-hagen:
I think I edited my comment accordingly just as you commented!  The tests should run fine now.

karllark:
Setting up a PR to v1.4 is ok, but in the future can you do the PR to master and then comment that it needs to be done for v1.4?  No change needed for this PR, just that is the way that makes things the most straightforward for me.  Thanks.

lea-hagen:
> Setting up a PR to v1.4 is ok, but in the future can you do the PR to master and then comment that it needs to be done for v1.4? No change needed for this PR, just that is the way that makes things the most straightforward for me. Thanks.

I'm happy to do that!  However, for some reason, when I tried the merge to master, it listed a whole bunch of commits and lots of line changes.  Perhaps that's because I made the branch from v1.4 instead of master?  I'll try branching from master next time.

karllark:
Good point.  Open the PR to whichever branch makes sense.  I can see that fixing v1.4 could mean you really do want to branch from v1.4.  It will get fun merging a fix into both at some point I imagine.

PR #573, opened 2020-06-03
Fix travis errors

karllark:
Leftover reference to an eztables function.
Add stub datamodel.py files to allow test collection for travis.  These stub files should be removed once #320 is implemented.


lea-hagen:
It looks like the `datamodel` stubs are breaking things.  When running the beast, it imports the one in the beast directories rather than importing the one in the user's directory.

PR #580, opened 2020-06-16
Fix a bug in pick_models_toothpick_style

### LINKED ISSUE #560: idxs referenced in pick_models_toothpick_style before assignment
There doesn't seem to be any code assigning `idxs` for the general case in `make_ast_inpit_list.py` (it gets referenced in `line 159`).

It seems like the original line declaring the `idxs` was removed in the last commit 20 days ago. It originally said `idxs = np.where(modelsedgrid.grid["logL"] > -9)[0] ` on `line 136`.

I don't know why this was deleted though. Any help in fixing this would be much appreciated.

galaxyumi:
Fix the bug on idxs in 'pick_models_toothpick_style'. This should resolve the issue #560. 

But don't know how to submit the same PR request for the BEAST v1.4 though.

This line below automatically links the PR to the issue and will automatically close the issue when the PR is merged.
Closes #560.

### DIFF
--- a/beast/observationmodel/ast/make_ast_input_list.py
+++ b/beast/observationmodel/ast/make_ast_input_list.py
@@ -131,7 +131,9 @@ def pick_models_toothpick_style(
     modelsedgrid = SEDGrid(sedgrid_fname)
 
     sedsMags = -2.5 * np.log10(modelsedgrid.seds[:] / vega_flux)
+    Nseds = sedsMags.shape[0]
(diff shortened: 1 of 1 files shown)


karllark:
For designating this PR as for v1.4 as well as the master, you state it just as you did.  I then add the v1.4 milestone and this tells me we need it for both.  That is the theory.  We'll see if it works.  ;-)

PR #582, opened 2020-06-16
Migrate to new packaging and CI tests

karllark:
Migrate to the new packaging (see https://docs.astropy.org/projects/package-template/en/latest/ape17.html).  This removes astropy-helpers and uses a cleaner test/build system (tox).

Move from travis CI to github actions for testing.  Faster and better integration.
Move from coveralls to codecov for coverage reporting.  A bit nicer interface (shows coverage and size of files at the same time graphically.


karllark:
To get the github actions to trigger, need to merge into master.  Wish it would trigger the first time on a PR, but that seems to not happen.  Will fix any issues in a subsequent PR.

karllark:
Tests should work - have spot checked via the command line.  The codestyle check will fail.  There are issues that need fixing.

PR #583, opened 2020-06-16
adding badges, removing files

karllark:


karllark:
Most things work.  Merging.

PR #579, opened 2020-06-15
beast_settings (previously datamodel) class

lea-hagen:
This is an updated version of #326 that implements `datamodel` as a class (with a text file of input parameters) rather than using a `datamodel.py` file that gets imported.  

Here's an example of how it works, copied over from #326.  I'll open a PR in beast-examples with the datamodel.txt file, but it's nearly identical to datamodel.py.
```
> from beast.tools import read_datamodel
> datamodel = read_datamodel.datamodel('beast-examples/phat_small/datamodel.txt')
> 
> datamodel.surveyname
'PHAT'
> datamodel.basefilters
['F275W', 'F336W', 'F475W', 'F814W', 'F110W', 'F160W']
> datamodel.obs_colnames
['f275w_rate', 'f336w_rate', 'f475w_rate', 'f814w_rate', 'f110w_rate', 'f160w_rate']
> datamodel.distance_unit
Unit("mag")
> datamodel.velocity
<Quantity -300. km / s>
> datamodel.osl
CompositeStellib, <beast.physicsmodel.stars.stellib.CompositeStellib object at 0x7f173f9f98d0>
['Tlusty', 'Kurucz 2004']
> datamodel.av_prior_model
{'name': 'flat'}
```
When reading in the file, it automatically runs the verification code (`beast/tools/verify_params.py`).  The user can always re-run the verification if they change anything:
```
> datamodel.verify_datamodel()
```

Still to do:
* finish testing that everything works as expected
* update docs

closes #320 
closes #576
closes #508
closes #587

Issue #566, opened 2020-05-21
pdf2d tests for test_fit_grid.py


karllark:
Would be good to add testing the pdf2d creation as part of the test_fit_grid.py test.
This emerged from the review of PR #553.

Issue #543, opened 2020-04-18
Fix insecure file location with tests


karllark:
Using /tmp as a place to store files can be insecure.  Causing a warning in Codacy.
Should be fixable using tempfile.NamedTemporaryFile from the python tempfile module.

PR #589, opened 2020-06-22
Move regression tests to a class

### LINKED ISSUE #565: Improve remote-data tests
Use tempfile to manage the files.  Has the nice feature of removing them once things are done.

Use pytest fixtures to download files only once and use in multiple tests.  Might multiple tests in separate files are merged into a single file, but should speed up the tests by only downloading once.

### LINKED ISSUE #588: add a way for tox to download library files
Check if there's a way to include the `beast get_libfiles` command when running tox, so that the tests can access the library files.

@karllark

karllark:
Moving most of the regression tests to a single file so they are in a single class.  This allows for files to be downloaded *once* for all the tests instead of over and over again for different tests.  This saves in download time and bandwidth and makes the tests faster to run.  The BEAST library files are downloaded using the tools.get_libfiles and so they no longer have to be passed as they are installed in the standard location (shown as possible by PR #579).

Tests on my laptop show that the reorganized tests take 250 sec and consume 1 Gb.  The original test organization results in the tests taking 500 sec and consuming 2 Gb.

Closes #588 
Closes #565
Supports #566
Partially addresses #543


karllark:
@lea-hagen : here is the rework of the regression tests.  What do you think?  How will this interact with the tests you are proposing in #559?

This should make it easier to switch between versions of the cached files based on the version of the BEAST.  Not sure how to do this yet, but I know it will be easier as the names of most of the files used are now in one place.  There was much more duplication of downloads that I had realized.

karllark:
@eteq : If you have time, could you have a look at the new regression check test file that uses a class to avoid repeat downloads of the same file?  I tried to do this with pytest fixtures, but couldn't figure out how to get it to work.  No worries if you don't have time, this is already a major improvement.

karllark:
> Something I just changed in #559 yesterday to make tests simpler is make a function called `beast_settings_mockup` that creates the `beast_settings` object with all the settings for phat_small.

I suggest that this function could be replaced by saving the beast_settings.txt file in the tools/data directory.  This be more consistent with how we are handling small text files.

karllark:
> One other minor comment - it would be helpful to add some docstrings to the functions to clarify which things they're testing. I've been staring at these a while, so I mostly know what they're doing, but that might not be the case for someone new.

Very good point.  I'll work on the docstrings.

karllark:
> Once that function is run once, the settings object can be used for various function inputs, and it can be edited as needed to change settings (e.g., in #559, change `n_subgrids` from 1 to 2). I think that would be useful here, since there are lots of places that parameters are copied into function calls.

Good point.

lea-hagen:
> > Something I just changed in #559 yesterday to make tests simpler is make a function called `beast_settings_mockup` that creates the `beast_settings` object with all the settings for phat_small.
> 
> I suggest that this function could be replaced by saving the beast_settings.txt file in the tools/data directory. This be more consistent with how we are handling small text files.

I did a function to return the beast_settings object so that each individual test wouldn't have to read in a text file.  But with everything in one class, that's no longer an issue!  So yes, making it a text file sounds good.  Though since it's only used for tests, should it be in the tests directory instead?  And which of us should make that change in which PR in which order?

karllark:
Yes, who should do this.  I would propose that I finish this PR w/o doing anything about beast_settings.  Then you could do a PR to update the beast_settings as discussed.  How does this sound?  Not 100% how this interacts with PR #559.  Maybe that PR will need to be closed and a new one created based on the consolidated regression tests?

lea-hagen:
That sounds good.  The test for create_physicsmodel will be straightforward to copy-paste into this class, and probably making a new branch will be the easiest way to do it.

karllark:
I'll update the docstrings this afternoon and then will merge this PR.  Then you can do your magic!

PR #597, opened 2020-06-29
removed noisegrid.close()

christinawlindberg:
Makes plot_completeness.py work again.

### DIFF
--- a/beast/plotting/plot_completeness.py
+++ b/beast/plotting/plot_completeness.py
@@ -67,8 +67,6 @@ def plot_completeness(
         noisegrid = noisemodel.get_noisemodelcat(str(noise_model))
         # get the completeness
         model_compl = noisegrid["completeness"]
-        # close the file to save memory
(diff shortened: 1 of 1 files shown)


christinawlindberg:
Can I ask why 2 of the checks failed and if it matters?

karllark:
Great question.

The failed checked are unrelated to your change.  These are failing as we are in the middle of  updating the cached files that we use for regression tests.  We have made some *minor* changes in the BEAST that means the cached versions of the files are no longer correct.  Hopefully we can get this all sorted out soon.  In addition, there are some remaining codestyle errors that we need to fix.  We've had these errors for a long time, but now they qualify as failing so that we are motivated to get them fixed (in the past we just ignored this failure).

christinawlindberg:
Ah, awesome to hear. I was worried deleting the line might have messed up a unit test or something like that. Thanks for the clarification!

PR #601, opened 2020-07-03
Add tool to calculate observation depths

lea-hagen:
This adds the tool `calculate_depth`, which will find either the flux or Vega mag corresponding to the requested completeness(es).  If the ASTs (and therefore the noise model) don't cover a given completeness, it returns a `NaN`.  Docs and a test are included.

@karllark I included an input for `vega_fname`, since I saw it in some other code that said it was useful for testing, but I'm not actually sure if it's necessary.

```
>>> from beast.tools import calculate_depth
>>> calculate_depth.calculate_depth(
          'beast_example_phat_seds.grid.hd5',
          'beast_example_phat_noisemodel.grid.hd5',
          completeness_value=[0.5, 0.75],
          vega_mag=True
    )
{"HST_WFC3_F275W": [25.000309202589012, 24.80610510139205],
    "HST_WFC3_F336W": [24.65974845352875, 24.338061586936263],
    "HST_ACS_WFC_F475W": [np.nan, np.nan],
    "HST_ACS_WFC_F814W": [np.nan, 24.368742437736692],
    "HST_WFC3_F110W": [np.nan, np.nan],
    "HST_WFC3_F160W": [21.99298441116123, 21.504534701422067]}
```

### DIFF
--- a/beast/tests/test_regresscheck.py
+++ b/beast/tests/test_regresscheck.py
@@ -33,6 +33,7 @@ from beast.fitting import fit
 from beast.tools import (
     get_libfiles,
     beast_settings,
+    calculate_depth,
     subgridding_tools,
     star_type_probability,
 )
@@ -848,6 +849,28 @@ class TestRegressionSuite:
 
         # compare to new table
         compare_tables(expected_star_prob, Table(star_prob))
+
+    def test_calculate_depth(self):
+        """
+        Test for calculate_depth.py
+        """
+        # calculate depth for 50% and 75% completeness
+        depth = calculate_depth.calculate_depth(
+            self.seds_fname_cache,
+            self.noise_fname_cache,
+            completeness_value=[0.5, 0.75],
+        )
+        # expected results
+        expected_dict = {
+            "HST_WFC3_F275W": [25.000309202589012, 24.80610510139205],
+            "HST_WFC3_F336W": [24.65974845352875, 24.338061586936263],
+            "HST_ACS_WFC_F475W": [np.nan, np.nan],
+            "HST_ACS_WFC_F814W": [np.nan, 24.368742437736692],
+            "HST_WFC3_F110W": [np.nan, np.nan],
+            "HST_WFC3_F160W": [21.99298441116123, 21.504534701422067],
+        }
+        # compare them
+        compare_tables(Table(expected_dict), Table(depth))
 
     # ###################################################################
     # tools.run tests
--- a/beast/tests/test_regresscheck.py
+++ b/beast/tests/test_regresscheck.py
@@ -33,7 +33,7 @@ from beast.fitting import fit
 from beast.tools import (
     get_libfiles,
     beast_settings,
-    calculate_depth,
+    calc_depth_from_completeness,
     subgridding_tools,
     star_type_probability,
 )
@@ -850,15 +850,16 @@ class TestRegressionSuite:
         # compare to new table
         compare_tables(expected_star_prob, Table(star_prob))
 
-    def test_calculate_depth(self):
+    def test_calc_depth_from_completeness(self):
         """
         Test for calculate_depth.py
         """
         # calculate depth for 50% and 75% completeness
-        depth = calculate_depth.calculate_depth(
+        depth = calc_depth_from_completeness.calc_depth(
             self.seds_fname_cache,
             self.noise_fname_cache,
             completeness_value=[0.5, 0.75],
+            vega_mag=True,
         )
         # expected results
         expected_dict = {
--- a/beast/tools/calculate_depth.py
+++ b/beast/tools/calculate_depth.py
@@ -1,0 +1,132 @@
+import numpy as np
+from astropy.table import Table, vstack
+from beast.physicsmodel.grid import SEDGrid
+import beast.observationmodel.noisemodel.generic_noisemodel as noisemodel
+from beast.observationmodel.vega import Vega
+
+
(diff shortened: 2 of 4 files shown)


beast/tools/calculate_depth.py
```diff
@@ -0,0 +1,132 @@
+import numpy as np
+from astropy.table import Table, vstack
+from beast.physicsmodel.grid import SEDGrid
+import beast.observationmodel.noisemodel.generic_noisemodel as noisemodel
+from beast.observationmodel.vega import Vega
+
+
+def calculate_depth(
```
karllark, line 8: Would there be a more descriptive name possible?  Something that includes the completeness or something like that?  Just would be better to have the name a bit more descriptive.

beast/tools/calculate_depth.py
```diff
@@ -0,0 +1,132 @@
...
+            first_ind = np.where(sort_comp > compl)[0][0]
+            # corresponding flux
+            comp_flux = sort_sed[first_ind]
+            # save it
+            if vega_mag:
+                depth_dict[filt].append(-2.5 * np.log10(comp_flux/vega_flux[f]))
+            else:
+                depth_dict[filt].append(comp_flux)
+            if 'F475W' in filt:
+                import pdb; pdb.set_trace()
```
karllark, line 129: Why a debugger statement?  Is this leftover from the development?

lea-hagen: oops, yep.

beast/tools/calculate_depth.py
lea-hagen: Can do!

lea-hagen:
Thanks for reviewing!  I updated the name to `calc_depth_from_completeness` and made the codestyle fixes.  Hopefully good to go now.

beast/tests/test_regresscheck.py
```diff
@@ -849,6 +850,29 @@ def test_star_type_probability_no_Av(self):
...
+            completeness_value=[0.5, 0.75],
+            vega_mag=True,
+        )
+        # expected results
+        expected_dict = {
+            "HST_WFC3_F275W": [25.000309202589012, 24.80610510139205],
+            "HST_WFC3_F336W": [24.65974845352875, 24.338061586936263],
+            "HST_ACS_WFC_F475W": [np.nan, np.nan],
+            "HST_ACS_WFC_F814W": [np.nan, 24.368742437736692],
+            "HST_WFC3_F110W": [np.nan, np.nan],
```
karllark, line 870: Why are these `nan` values?  Is there something wrong with the ASTs in the phat_small example?  Mainly curious.

docs/other_tools.rst
```diff
@@ -23,3 +24,32 @@ and save it to disk:
...
+          completeness_value=[0.5, 0.75],
+          vega_mag=True
+      )
+  >>> # Depth in F275W (Vega mag)
+  >>> depth['HST_WFC3_F275W'] #doctest: +SKIP
+  [25.000309202589012, 24.80610510139205]
+  >>>
+  >>> # Depth in F814W (Vega mag)
+  >>> # NaNs show that ASTs don't go deep enough to evaluate 50% completeness
+  >>> depth['HST_ACS_WFC_F814W'] #doctest: +SKIP
```
karllark, line 54: Are the `>>>` needed?  I would have thought not as you already have declared this as a code block.  Without the `>>>` characters, then the `#doctest: +SKIP` shouldn't be needed.

beast/tests/test_regresscheck.py
lea-hagen: As far as I've figured out, there are two reasons for nans.  One is if the ASTs don't cover the whole flux range of the grid, so there aren't statistics for faint sources.  The other is if the observations are so deep that even the faintest sources are recovered at some non-zero rate.

lea-hagen: Here's an example for one of the METAL fields, using completenesses of 0.2, 0.25, 0.3:
```
{'HST_WFC3_F225W': [42.48340001978902, 36.85778870609744, 36.54311413627377],
'HST_WFC3_F275W': [nan, nan, 37.442233109245905],
'HST_WFC3_F336W': [nan, 38.19925138344242, 37.2154373318946],
'HST_WFC3_F475W': [nan, nan, 33.83419535967331],
'HST_WFC3_F814W': [28.871186766091327, 28.685084890012547, 28.526476367397507],
'HST_WFC3_F110W': [26.802812634408944, 26.66268095677285, 26.538354275870823],
'HST_WFC3_F160W': [25.54145798073774, 25.387076912540735, 25.252994542836568]}
```
What I think this is saying (and please correct me if I'm wrong) is that for the faintest blue sources, they're bright enough in another filter to still be recovered ~25% of the time.

docs/other_tools.rst
lea-hagen: I put them there to distinguish between code that's typed in the console and the printed outputs from the code.

beast/tests/test_regresscheck.py
karllark: Ok.  Thanks.  This just emphasizes that it would be great to get a better set of observations/ASTs for the regression tests and example run.  But that is a topic for another PR.

docs/other_tools.rst
karllark: This is ok.  Just supposes that this in done at the command prompt and not in code.  We can always change this later if it makes sense.

PR #602, opened 2020-07-15
adding input flux/rate/vega to simulation output table

karllark:
Adds in the input flux/rate/vega(mag) to the simulation output.  

This supports the use of the simulations as a way to interpolate the ASTs to generate a larger list of ASTs (useful for MATCH).

### DIFF
--- a/beast/observationmodel/observations.py
+++ b/beast/observationmodel/observations.py
@@ -293,12 +293,10 @@ def gen_SimObs_from_sedgrid(
 
         bname = filter.split(sep="_")[-1].upper()
         fluxname = f"{bname}_FLUX"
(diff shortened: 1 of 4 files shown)


karllark:
> Overall looks great! The only thing I'm wondering is about column names in the output table. For instance, it's not immediately obvious what the difference is between a column named `F475W_INPUT_FLUX` and `F475W_FLUX` without digging into the code. Perhaps add a list of column descriptions to the docs?

I did.  Maybe it needs to be even clearer?  An example maybe instead of `band_flux` and `band_input_flux`?

lea-hagen:
Oh wow, I totally missed that in the docs!  In that case, everything looks good.

PR #603, opened 2020-07-16
fix plot_cmd.py savefig keyword

lea-hagen:
While chatting with @kmcquinn about `plot_cmd`, we ran into a problem with the `savefig` keyword.  Renaming it to `save_fig` solves the problem - apparently argparse was somehow convinced that the keyword was already defined elsewhere.  I also clarified the usage of the keyword in the docstring and in the docs.

### DIFF
--- a/beast/plotting/plot_cmd.py
+++ b/beast/plotting/plot_cmd.py
@@ -19,7 +19,7 @@ def plot_cmd(
     mag1_filter="F475W",
     mag2_filter="F814W",
     mag3_filter="F475W",
-    savefig=False,
+    savefig=None,
     show_plot=True,
 ):
     """
@@ -28,17 +28,17 @@ def plot_cmd(
 
     Parameters
     ----------
-    fitsfile:           str
+    fitsfile : str
         input fitsfile (includes full path to file); format = .fits
-    mag1_filter:        str
-        1st color filter; default = 'F475W'
-    mag2_filter:        str
-        2nd color filter; default = 'F814W'
-    mag3_filter:        str
-        magnitude; default = 'F475W'
-    savefig:            boolean
-        save figure; default = False
-    show_plot:          boolean
+    mag1_filter : str (default='F475W')
+        1st color filter
+    mag2_filter : str (default='F814W')
+        2nd color filter
+    mag3_filter : str (default='F475W')
+        magnitude
+    savefig : str or None (default=None)
+        to save the figure, set this to the file extension (e.g., 'png', 'pdf')
+    show_plot : boolean
         True, show the plot (to screen or a file)
         False, return the fig
     """
@@ -111,7 +111,10 @@ if __name__ == "__main__":  # pragma: no cover
         help="Choose filter for the magnitude",
     )
     parser.add_argument(
-        "--savefig", action="store", default="True", help="Save figure or plot it",
+        "--savefig",
+        action="store",
+        default=None,
+        help="To save the figure, set this to the file extension (e.g., 'png', 'pdf')",
     )
 
     args = parser.parse_args()
--- a/beast/plotting/plot_cmd.py
+++ b/beast/plotting/plot_cmd.py
@@ -19,7 +19,7 @@ def plot_cmd(
     mag1_filter="F475W",
     mag2_filter="F814W",
     mag3_filter="F475W",
-    save_fig=None,
+    savefig=None,
     show_plot=True,
 ):
     """
@@ -36,7 +36,7 @@ def plot_cmd(
         2nd color filter
     mag3_filter : str (default='F475W')
         magnitude
-    save_fig : str or None (default=None)
+    savefig : str or None (default=None)
         to save the figure, set this to the file extension (e.g., 'png', 'pdf')
     show_plot : boolean
         True, show the plot (to screen or a file)
@@ -80,8 +80,8 @@ def plot_cmd(
 
     # save or show fig
     if show_plot:
-        if save_fig:
-            fig.savefig("{}.{}".format(basename, save_fig))
+        if savefig:
+            fig.savefig("{}.{}".format(basename, savefig))
         else:
             plt.show()
     else:
@@ -110,12 +110,6 @@ if __name__ == "__main__":  # pragma: no cover
         default="F475W",
         help="Choose filter for the magnitude",
     )
-    parser.add_argument(
-        "--save_fig",
-        action="store",
-        default=None,
-        help="To save the figure, set this to the file extension (e.g., 'png', 'pdf')",
-    )
 
     args = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


lea-hagen:
Looks like the test is failing because of changes to the simobs reference files associated with #602, so should be fine after that's merged.

karllark:
Are you using the "master" script way to run this script?  Like `beast plot_cm` or another way?  I'm curious to know why this doesn't work with `savefig`, but does with `save_fig`.  Very odd behavior.

lea-hagen:
Here's the traceback I got with the `savefig` keyword.  It didn't even get as far as printing the help message!
```
(bdev) [lhagen@plhavarti beast_lea-hagen] python -m beast.plotting.plot_cmd
Traceback (most recent call last):
  File "/user/lhagen/miniconda3/envs/bdev/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/user/lhagen/miniconda3/envs/bdev/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/astro/dust_kg3/lhagen/stsci/beast_lea-hagen/beast/plotting/plot_cmd.py", line 117, in <module>
    help="To save the figure, set this to the file extension (e.g., 'png', 'pdf')",
  File "/user/lhagen/miniconda3/envs/bdev/lib/python3.6/argparse.py", line 1352, in add_argument
    return self._add_action(action)
  File "/user/lhagen/miniconda3/envs/bdev/lib/python3.6/argparse.py", line 1715, in _add_action
    self._optionals._add_action(action)
  File "/user/lhagen/miniconda3/envs/bdev/lib/python3.6/argparse.py", line 1556, in _add_action
    action = super(_ArgumentGroup, self)._add_action(action)
  File "/user/lhagen/miniconda3/envs/bdev/lib/python3.6/argparse.py", line 1366, in _add_action
    self._check_conflict(action)
  File "/user/lhagen/miniconda3/envs/bdev/lib/python3.6/argparse.py", line 1505, in _check_conflict
    conflict_handler(action, confl_optionals)
  File "/user/lhagen/miniconda3/envs/bdev/lib/python3.6/argparse.py", line 1514, in _handle_conflict_error
    raise ArgumentError(action, message % conflict_string)
argparse.ArgumentError: argument --savefig: conflicting option string: --savefig
```

karllark:
What do you get if you run `beast plot_cmd` from anywhere on your system where you've got the beast installed?

lea-hagen:
Yes, `beast plot_cmd` does work.  Very puzzling!

karllark:
Yes, this is.  I might try to investigate before merging this PR.  We use savefig for a *bunch* of plotting scripts.  Would be nice to not have to change them all.

karllark:
I found it!  It is in initialize_parser.  This already defines the savefig argument.  So, the right thing to do is to remove the savefig argument from the plot_cmd main block.  --savefig will then work as expected.

lea-hagen:
I think I tracked it down: `plot_cmd` runs 
```
from beast.plotting.beastplotlib import initialize_parser
```
That parser has a `savefig` option.

karllark:
Great minds!

lea-hagen:
Hahaha, excellent!  I'll update it to do that instead.

lea-hagen:
It looks like the reason `beast plot_cmd` worked is because `main.py` extracts the inputs for the `plot_cmd` function (within `plot_cmd.py`) and lists those, rather than going though the argparse interface.  Doing the help entry for `plot_cmd` does this:
```
(bdev) [lhagen@plhavarti beast_lea-hagen] python -m beast.plotting.plot_cmd -h
usage: plot_cmd.py [-h]
                   [-s {png,jpg,jpeg,pdf,ps,eps,rgba,svg,tiff,tif,pgf,svgz,raw}]
                   [-t] [--mag1 MAG1] [--mag2 MAG2] [--mag3 MAG3]
                   filename

positional arguments:
  filename              Path to FITS file to plot

optional arguments:
  -h, --help            show this help message and exit
  -s {png,jpg,jpeg,pdf,ps,eps,rgba,svg,tiff,tif,pgf,svgz,raw}, --savefig {png,jpg,jpeg,pdf,ps,eps,rgba,svg,tiff,tif,pgf,svgz,raw}
                        Save figure to a file of specified type rather than
                        show it. Must be one of: "png", "jpg", "jpeg", "pdf",
                        "ps", "eps", "rgba", "svg", "tiff", "tif", "pgf",
                        "svgz", "raw"
  -t, --tex             Configure Matplotlib to use LaTeX; sets rcParam
                        "usetex" to True. (Requires working TeX installation.)
                        Defaults to false.
  --mag1 MAG1           Choose filter for mag1 (color=mag1-mag2)
  --mag2 MAG2           Choose filter for mag2 (color=mag1-mag2)
  --mag3 MAG3           Choose filter for the magnitude
```
Whereas the help with the `beast` interface gives the names in the function:
```
usage: beast plot_cmd [-h] [--mag1_filter MAG1_FILTER]
                      [--mag2_filter MAG2_FILTER] [--mag3_filter MAG3_FILTER]
                      [--savefig SAVEFIG] [--show_plot SHOW_PLOT]
                      fitsfile

positional arguments:
  fitsfile

optional arguments:
  -h, --help            show this help message and exit
  --mag1_filter MAG1_FILTER
  --mag2_filter MAG2_FILTER
  --mag3_filter MAG3_FILTER
  --savefig SAVEFIG
  --show_plot SHOW_PLOT
```

karllark:
Yep.  There is a different in the args between the two methods of running.  Haven't get adjusted to the differences that are imposed by having the `beast plot_cmd` generate the commandline args based on the function arguments.  If we write the functions with this in mind, we might even get away with no `main` in each script file.  It would be better to use the `beast plot_cmd` interface as this works anywhere and is not as cumbersome as the `python -m` method.

lea-hagen:
That makes sense.  I agree that the `python -m` is suboptimal.  Is there a way for the `beast [function]` method to show info from the docstring?  (though that's a bit beyond the scope of this PR!)

And thanks for restarting the tests!

karllark:
Probably there is a way to use the docstring.  Check out the beast/main.py code.   Maybe a clue there?  Can you make an issue for this?

beast/plotting/plot_cmd.py
```diff
@@ -19,7 +19,7 @@ def plot_cmd(
     mag1_filter="F475W",
     mag2_filter="F814W",
     mag3_filter="F475W",
-    savefig=False,
+    savefig=None,
```
karllark, line 22: Why this change?

lea-hagen: Ah, the original implementation had the argparse default to True, so it was trying to set True as the file extension (which obviously didn't work).  I can switch this back to a boolean if that's preferred.

karllark: I think that is the right way to go.  May not matter, but it was working with False before.

PR #607, opened 2020-07-28
Fix attribute name for settings object

lea-hagen:
I stumbled across a couple places with an incorrect attribute name (should be `settings.settings_file`, not `settings.input_settings_file`).


lea-hagen:
I tried force-pushing in case an out-of-date master was to blame for failing tests.  But I just realized it's from #605.

karllark:
> I tried force-pushing in case an out-of-date master was to blame for failing tests. But I just realized it's from #605.

Yep.  I've been working on getting the tests to pass for that PR.  We definitely need to create an automatic way to regenerate all the regression test files when we change something in the "core".  For this PR, that's updating the observation model calculation.

lea-hagen:
Agreed.  I just opened an issue 😄

Issue #598, opened 2020-06-30
Regression test failures


karllark:
We have updated things in the BEAST that are causing some of the regression tests to fail.  The cached files used by these tests need updating.  Most of the standard cached files (i.e., generated by running the phat_small example) have already been updated.   The remaining regression tests that need updated files or revising so that they are independent of the cached files are:

- [ ] test_star_type_probability_all_params
- [ ] test_compare_spec_type_inFOV
- [ ] test_ast_pick_models
- [ ] test_read_lnp_data
- [ ] test_read_noise_data

lea-hagen:
I'm pretty sure the underlying reason for the failures of these is the number of PDF bins used in the fitting.  Going from 50 to 100 bins changes the 1D/2D PDFs (obviously) as well as the stats.  I keep getting mixed up with how many bins the tests should be using, so it would help to clarify that.

karllark:
I don't think it is the number of bins, but that the SED file has changed due to removing the LogL < -9. points from the isochrones.

lea-hagen:
Oh yeah, that too!

karllark:
These tests are marked to be skipped in PR #596.  When each test is fixed, the skip notation should be removed.   Skipping the tests allows for the coverage report to be generated which is useful.

Issue #567, opened 2020-05-21
Testing w/ cached files support different versions


karllark:
Need to update the testing to support different versions of the cached files.  Required to ensure the v1.4 and v2.0 branches are both working.  Bug fixes for v1.4 in particular.

karllark:
Done in PR #612.

PR #612, opened 2020-09-03
Update the regression test to METAL observed/AST dataset

### LINKED ISSUE #325: New observed data/ASTs for regression test
The observed data used for the regression test could use updating.   Currently, PHAT photometry from the deep past is being used.  The pedigree of this data is not clear.  It works for what we need, but it would be better to have a small observed dataset whose origins we clearly document.  Same for the ASTs that are used as part of the testing.

karllark:
The current regression tests are based on the beast-examples/phat_small dataset (obs/AST results).  The details of the phat_small dataset are lost in the mists of time.  The phat_small dataset is old enough that it does not have the full information that we now use in the BEAST (e.g., the ASTs only report mags for the recovered measurements, instead of fluxes that all more recent AST results provide).

This new dataset is based on a subset of a METAL field.

A number of regression tests are skipped as they need updated cache files either for phat or metal or both.  See #598.

Includes minor updates to plot_indiv_fit.py.
Includes various updates to close files left open and remove some deprecation warnings.   This cleans up the test output making it easier to see problems.

Partially addresses #567.
Closes #325.

### DIFF
--- a/beast/plotting/plot_cmd_with_fits.py
+++ b/beast/plotting/plot_cmd_with_fits.py
@@ -60,8 +60,10 @@ def plot(
     # read in data
     data_hdu = fits.open(data_fits_file)
     data_table = data_hdu[1].data
+    data_hdu.close()
     beast_hdu = fits.open(beast_fits_file)
     beast_table = beast_hdu[1].data
+    beast_hdu.close()
 
     # figure out the subset that were modeled
     data_cat = SkyCoord(
--- a/beast/plotting/plot_noisemodel.py
+++ b/beast/plotting/plot_noisemodel.py
@@ -151,7 +151,7 @@ if __name__ == "__main__":  # pragma: no cover
         help="path+name of the noise model file(s)",
     )
     parser.add_argument(
-        "phot_file", type=str, help="name of the file to save the plot",
+        "plot_file", type=str, help="name of the file to save the plot",
     )
     parser.add_argument(
         "--samp", type=int, default=100, help="plot every Nth point",
--- a/beast/plotting/plot_noisemodel.py
+++ b/beast/plotting/plot_noisemodel.py
@@ -176,7 +176,7 @@ if __name__ == "__main__":  # pragma: no cover
     plot_noisemodel(
         args.sed_file,
         args.noise_file_list,
-        args.phot_file,
+        args.plot_file,
         samp=args.samp,
         color=args.color,
         label=args.label,
--- a/beast/tests/helpers.py
+++ b/beast/tests/helpers.py
@@ -5,6 +5,7 @@ import os.path
 import numpy as np
 import h5py
 import tempfile
+import requests
 
 from astropy.io import fits
 from astropy.utils.data import download_file
@@ -23,12 +24,20 @@ def download_rename(filename, tmpdir=""):
         name of file to download
     """
     url_loc = "http://www.stsci.edu/~kgordon/beast/"
+    #    url = f"{url_loc}{filename}"
+    # check if the file exists, return False if it doesn't
+    #    response = requests.head(url)
+    #    if response.status_code in [301, 302, 303, 307, 308, 200]:
     fname_dld = download_file("%s%s" % (url_loc, filename))
     extension = filename.split(".")[-1]
     # fname = f"{fname_dld}.{extension}"
     fname = tempfile.NamedTemporaryFile(suffix=f".{extension}").name
     os.rename(fname_dld, fname)
     return fname
+
+
+#    else:
+#        return False
 
 
 def compare_tables(table_cache, table_new, rtol=1e-7, otag=""):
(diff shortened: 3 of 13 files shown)


beast/plotting/plot_cmd_with_fits.py
```diff
@@ -60,8 +60,10 @@ def plot(
     # read in data
     data_hdu = fits.open(data_fits_file)
     data_table = data_hdu[1].data
+    data_hdu.close()
     beast_hdu = fits.open(beast_fits_file)
     beast_table = beast_hdu[1].data
+    beast_hdu.close()
```
lea-hagen, line 66: Thanks for catching that these hadn't been closed!  Would it be worth using `with fits.open() as...` instead?

beast/plotting/plot_noisemodel.py
```diff
@@ -151,7 +151,7 @@ def plot_noisemodel(
         help="path+name of the noise model file(s)",
     )
     parser.add_argument(
-        "phot_file", type=str, help="name of the file to save the plot",
+        "plot_file", type=str, help="name of the file to save the plot",
```
lea-hagen, line 154: It looks like this is supposed to be `phot_file`.  Was this accidentally changed?  Or maybe the `args.phot_file` below needs to be updated too.

beast/tests/helpers.py
```diff
@@ -23,6 +24,10 @@ def download_rename(filename, tmpdir=""):
         name of file to download
     """
     url_loc = "http://www.stsci.edu/~kgordon/beast/"
+    #    url = f"{url_loc}{filename}"
+    # check if the file exists, return False if it doesn't
+    #    response = requests.head(url)
+    #    if response.status_code in [301, 302, 303, 307, 308, 200]:
```
lea-hagen, line 30: Do these still need to be here?

beast/tests/test_regresscheck.py
```diff
@@ -326,24 +350,26 @@ def test_fit_grid(self):
 
     # ###################################################################
     # AST tests
-    @pytest.mark.skip(reason="updated cached file needed")
+    @pytest.mark.skip(reason="need filters info: get from sed grid?")
```
lea-hagen, line 353: Could you clarify what's needed here?  I'm surprised it's not working.

beast/tools/tests/test_write_sbatch_file.py
```diff
@@ -20,6 +20,7 @@ def test_sbatch_file():
 
     file = open(temp_file.name)
     content = file.read()
+    file.close()
```
lea-hagen, line 23: Another place where `with open() as ...` may or may not be useful.

lea-hagen (changes requested):
Looks good!  I like the new file organizational system for the tests (both in the new class and in the folders online).

One comment about `plot_indiv_fit`: the filters are still hard-coded in, so it'll only work for a particular survey.  I'd imagine it would be straightforward to extract them from the stats file, and maybe use pysynphot or something to get the wavelength from the corresponding filter curve.  But that may be outside the scope of this PR :).

For the tests, perhaps getting everything consistently regenerated would be a good topic for an upcoming hack day??? 😄

beast/plotting/plot_cmd_with_fits.py
karllark: We could.  This works and is explicit.  Maybe better????  No real idea.

beast/plotting/plot_noisemodel.py
karllark: Nope, supposed to be *plot* as it is the file to save the plot to.  Having it as phot confused me.  Will fix the args variable name.  Good catch.

beast/tests/helpers.py
karllark: Nope.  Will delete.

beast/tests/test_regresscheck.py
karllark: Like all the tests - depends on if it is phat_small or metal_small set of cached files.  This test would work with either, but needs to know the filters.   The filters change between the two cached file sets.  This is a note to indicate this (not a great note - but at least a clue).  I figured a separate PR (or multiple PRs) to get the regression tests all working *and* setup for being able to have different cached sets.  This will be very useful in the future when we break backwards compatability.

beast/tools/tests/test_write_sbatch_file.py
karllark: I'm ok with this for now.

beast/tests/test_regresscheck.py
karllark: I've added a bit more info to make this comment more helpful.

karllark:
> One comment about `plot_indiv_fit`: the filters are still hard-coded in, so it'll only work for a particular survey. I'd imagine it would be straightforward to extract them from the stats file, and maybe use pysynphot or something to get the wavelength from the corresponding filter curve. But that may be outside the scope of this PR :).

Good point!  I've been wondering why plot_indiv_fit has been looking odd for the metal results.  Will see what I can do about this in the *next* PR (not this one).

karllark:
> For the tests, perhaps getting everything consistently regenerated would be a good topic for an upcoming hack day???

Good idea.  I've been thinking we should have one soon.  I'll setup a doodle poll.

lea-hagen (approved):
Thanks for making the changes!

PR #614, opened 2020-09-08
splitting catalogs by source density: add keyword for minimum number of sub-files

lea-hagen:
As I'm doing production runs, I've come across a problem in which source densities with a small number of stars don't get split into multiple subfiles, and therefore don't get trimmed much and take forever to fit.  So I added an optional keyword to `split_catalog_using_map`, `min_n_subfile`, which tells the minimum number of subfiles to make.

As an example, let's say that some field's SD bin 0-1 has 20,000 stars and bin 1-2 has 300 stars.  I've been using `n_per_file=1000`, which splits bin 0-1 into 20 subfiles and doesn't split bin 1-2.  This works great to make the fitting quick for bin 0-1, but bin 1-2 still takes a very long time, since it hasn't been broken up by flux.  Now, I can set `min_n_subfile=10`: bin 0-1 already has 20 subfiles, so nothing changes there, but now bin 1-2 gets sorted by flux and split into 10 subfiles.

I've updated the docs to add the new keyword.

### DIFF
--- a/beast/tools/split_catalog_using_map.py
+++ b/beast/tools/split_catalog_using_map.py
@@ -40,6 +40,14 @@ def main():  # pragma: no cover
     )
 
     parser.add_argument(
+        "--min_n_subfile",
+        type=int,
+        default=None,
+        help="""Set the minimum number of subfiles to use per bin (relevant if a
+        bin has fewer than n_per_file but you still want flux-sorted subfiles)""",
+    )
+
+    parser.add_argument(
         "--sort_col",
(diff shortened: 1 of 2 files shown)


karllark:
Assign me as a reviewer when this PR is ready to go.

lea-hagen:
Hmm, it looks like only owners and folks with write access are allowed to request reviewers.  So consider this my less-formal review request!

karllark (approved):
Looks good.

PR #618, opened 2020-09-09
Simulations based on age/mass prior model

### LINKED ISSUE #593: Absolute SFH for simulations/MegaBEAST
The BEAST only requires the relative SFH for the fitting.

The MegaBEAST will fit for the absolute SFH as part of its fitting.  The absolute SFH translates to the number of stars detected and the number of stars detected is needed to correct for the stochastic nature of sampling of the IMF, etc. to get correct fits when the number of detected sources is small (main use case for the MegaBEAST).

For BEAST simulations, the absolute SFH matters in that the *mass* of stars formed per year will give the *number* of stars formed in that year as sampled from the IMF.   So, need to add the absolute (or use them) SFH values in the simulations to have better simulations and support MegaBEAST.  This issue emerged from the MATCH/(Mega)BEAST comparison efforts.

In the current simulation code, the total number of stars desired in the simulation is set and then that number of stars is simulated based on the full "weight" of each bin in the BEAST multiplied by the completeness in a single filter (e.g., only the toothpick model supported at this time).  The weight is the product of the prior (e.g., relative SFH, IMF, extinction, etc.) and bin width.

Should consider using the maximum completeness across all the observed filters to do the simulations (instead of just a single filter).  This should(?) better mimic source detection that uses all bands.

karllark:
The number of stars in a simulation can now be computed from age/mass priors.  This is needed to get simulations that work for testing the MegaBEAST (and MATCH).  It is still possible to set the number of stars to simulate explicitly (so - two options now).

To provide the age/mass priors to the simulation code, a new file was created to save information on the BEAST model/run.  This is the "beast_info" file and is saved as as ASDF file as this provides a *very* easy way to store dictionaries.  This beast info file will be used by the MegaBEAST as well.

As part of this work, it was noticed that the plot created from `plot_indiv_plot` looked a bit odd - specifically for the lines/area for the stellar+dust+bias fluxes.  This was traced to the use of symlogs to handle the large dynamic range and negative values for these fluxes.  The symlog threshold (transition between log and linear regimes) was set much too high.  A threshold of 1e-20 was found to work much better.  A symlog module was added to provide the needed functions.  The flux plot in `plot_indiv_fit` was changed to be a symlog plot (supported by matplotlib!) to show the positive and negative fluxes (model and observed).  The default max number of bins in the 1D PDFs has been changed from 100 to 200 to better resolve the 1D PDFs - especially those based on symlogs.

Docs updated and tests included.

Closes #593.


kmcquinn:
**Review of PR by McQuinn:**
1) The output file gives the simulated band_vega as vega magnitudes with zero and negative fluxes given as -99.999. MATCH, however, uses 99.999 as the non-recovered mag values. I think having the minus sign will cause problems when fitting BEAST simulated photometry as a test of the SFH recovery process. So, if it doesn't cause issues with BEAST, it would be preferable to have the vega band mag with zero and negative fluxes listed as 99.999  (i.e., no minus sign). 

2) There are some leftover formatting/typos from the previous documentation:
This provides

the star formation rate verus time in the age prior combined with the

is given by the --nsim parameter.

3) Is there a sample beastinfo ASDF file that can be added to the documentation? What information does this file contain?

4) One option in the simulations is which filter to use for the completeness function:
" The filter to use for the completeness function is given by the --compl_filter parameter (default=F475W). Set compl_filter=max to use the max completeness value across all the filters. "

How does this impact the simulated photometry? Any guidance on how to choose which filter to use? Or is this similar to running the BEAST whereh F475W is just generally used?

Thanks!

beast/physicsmodel/prior_weights_stars.py
```diff
@@ -83,7 +87,8 @@ def compute_age_prior_weights(logages, age_prior_model):
         )
 
     # normalize to avoid numerical issues (too small or too large)
-    age_weights /= np.average(age_weights)
+    # do not normalize as the absolute level now matters for simulations and megaBEAST
+    # age_weights /= np.average(age_weights)
```
lea-hagen, line 91: Since normalization was originally to avoid numerical issues, do we need to worry about that now?

beast/plotting/plot_toothpick_details.py
```diff
@@ -60,6 +62,10 @@ def plot_toothpick_details(asts_filename, seds_filename, savefig=False):
 
         # not all bins are filled with good data
         ngbins = model._nasts[i]
+        print(model._minmax_asts[:, i])
+
+        print(model._fluxes[0:ngbins, i])
```
lea-hagen, line 67: Do these still need to be here?

lea-hagen (changes requested):
Looking good!  I have a few questions/comments.
* As best I can tell, the purpose of `beast_info` is to keep a record of the beast settings, so that later it's easy to find the settings that were used.  Is that correct?  If so, what is the advantage of transferring them to an asdf file rather than, say, running `shutil.copy2(beast_settings_file, project_dir)` and then using the existing beast_settings infrastructure?
* As `gen_SimObs_from_sedgrid` keeps getting more substantial (I think in the past it was maybe 20-30 lines?), I'm wondering if it should move to its own file.  That way `observations.py` doesn't get too unwieldy.
* Could you show an example of the new `plot_indiv_fit`?

docs/simulations.rst
```diff
@@ -2,27 +2,41 @@
...
+done by computing the total mass formed by integrating the star formation
+rate versus time (age prior) multiplied by the total mass formed per unit time
+(mass prior).  This number of total stars formed is then the total mass divided
+by the average mass computed from the mass prior model.  This total number of
+stars is simulated and then the completeness function is applied to remove
+all the simulated stars that would not be observed.  This provides
+
+the star formation rate verus time in the age prior combined with the
+
+is given by the `--nsim` parameter.
```
lea-hagen, line 38: Woops, forgot to add that this has a few extra line breaks!

lea-hagen: (which I also just realized was in @kmcquinn's review, too)

karllark:
Thanks for the review @kmcquinn.  Responses:

1) Done.

2) Fixed.

3) This file contains saved meta info from the run to better log what information that is needed for the simulations *and* by the MegaBEAST.  I expect the content of this file will expand significantly to as the MegaBEAST is fully developed.  Definitely should say something in the documentation.  I've added a line to the `outputs` docs.

4) I am not sure of the right answer here.  I'm also not sure that it makes a huge difference.  The reason is that the toothpick model is an approximation of what is really happening when the photometry is measured.  In reality, there is only one completeness value for a particular model star that depends on the flux of that model in *all* the bands (i.e., brightness and shape).  We aren't doing enough ASTs currently to define the completeness in that space.  Hence -> approximation using toothpick model.  My gut says that `max` is probably the right thing to use for the completeness, but I don't have any quantitative proof of this yet.

karllark:
> Looking good! I have a few questions/comments.
> 
> * As best I can tell, the purpose of `beast_info` is to keep a record of the beast settings, so that later it's easy to find the settings that were used.  Is that correct?  If so, what is the advantage of transferring them to an asdf file rather than, say, running `shutil.copy2(beast_settings_file, project_dir)` and then using the existing beast_settings infrastructure?

I did think about this.  But I thought that a separate output info file would be better as then we can put anything we want in it, not just what was passed in via the settings.  I already have done this by including not just the prior models, but also the 1D vectors of the actual grid values for the dust parameters. 

> * As `gen_SimObs_from_sedgrid` keeps getting more substantial (I think in the past it was maybe 20-30 lines?), I'm wondering if it should move to its own file.  That way `observations.py` doesn't get too unwieldy.

Could be done.  I guess my idea was that these are simulations of the observations so they belonged there.  We have lots of longer files, so I just left it there.  We should think about this more globally for all the BEAST files.  Moving things can break other scripts/code, so...

> * Could you show an example of the new `plot_indiv_fit`?

Yeah, good point.  I've changed the example to be from the metal_small example instead of phat_small.  phat_small is something we just shouldn't advertise anymore anyway.

beast/tools/symlog.py
```diff
@@ -0,0 +1,48 @@
...
+
+    C : float
+        value of the threshold where the symlog changes from linear to log
+
+    Returns
+    -------
+    y_symlog : float
+        symlog of y
+    """
+    return np.sign(y) * np.log10(1.0 + np.abs(y / C))
```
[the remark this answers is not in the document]
karllark: I'm not sure this is an issue anymore given that C is 1e-20.  We shouldn't see anything really small anymore.  This should be a separate issue so that the math can be checked.

beast/tools/symlog.py
```diff
@@ -0,0 +1,48 @@
...
+
+    C : float
+        value of the threshold where the symlog changes from linear to log
+
+    Returns
+    -------
+    y_linear : float
+        linear version of y
+    """
+    return np.sign(y) * C * (-1.0 + 10.0 ** np.abs(y))
```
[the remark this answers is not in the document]
karllark: Same reply as above.

beast/tools/tests/test_symlog.py
```diff
@@ -0,0 +1,26 @@
...
+        4.34272769e-05,
+        3.01029996e-01,
+        1.04139269e00,
+        1.00000000e01,
+    ]
+
+    # test going to symlog
+    np.allclose(symlog(y), exp_symlog)
+
+    # test converting back to liner
```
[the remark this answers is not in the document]
karllark: Fixed.

beast/physicsmodel/prior_weights_stars.py
karllark: I just did a blanket normalization to all the priors in multiple places.  I checked and it should not be an issue for any reasonable age prior.

docs/simulations.rst
karllark: Fixed.

karllark:
@lea-hagen & @kmcquinn : ok to merge this PR?  I.e., were my responses to your reviews sufficient?

kmcquinn:
@karllark Just want to clarify something about the beastinfo ASDF file.  Is this something the user creates? Or is this generated by BEAST based on the inputs to the simulation? If it's the latter, then I'm happy with the responses. Thank you!

karllark:
> @karllark Just want to clarify something about the beastinfo ASDF file. Is this something the user creates? Or is this generated by BEAST based on the inputs to the simulation? If it's the latter, then I'm happy with the responses. Thank you!

It is generated by the BEAST.  It is way for the BEAST to save info that is useful later.  Definitely something that a users should *not* create by hand.

beast/tools/symlog.py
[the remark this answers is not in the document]
lea-hagen: Fair point.  I just did a quick test with y=1e-30, and the difference in the two methods is 0.0000008%, so I agree that it's fine!

beast/plotting/plot_toothpick_details.py
karllark: Removed.

lea-hagen (approved):
Changes look great!

kmcquinn:
I think it looks good, too! I'm excited to try it out!

PR #624, opened 2020-09-21
beast graphical model docs

### LINKED ISSUE #623: BEAST graphic - useful for docs?
As a follow on of @eteq's great tutorial this morning, I've attempted to create a graphical model for the BEAST.  This is just the 1st cut at doing this.  I wonder if this would be useful to add to the docs?  Comments?  Questions?  Suggestions for improvements?

Both have been updated based on comments given below.  See PR #624 for notebook that creates these figures.

Here is the version using understandable words.

beast-graphic

Here is the version using symbols based on the equations in the BEAST paper.

beast-graphic-symbols

karllark:
Adding in code that creates the beast graphical model visualizations.  Figures added to docs.

Closes #623.


karllark:
should change this to a script - easy to support two different versions.

s-goldman (approved):
Looks good to me.

PR #625, opened 2020-09-24
fix keyword in simulation docs

lea-hagen:
The keyword for the new beast_info files in `simulate_obs` is `beastinfo_list`, but the docs had it as `beastinfo`, so I did a quick fix.


karllark (approved):
Good catch!

PR #635, opened 2020-10-02
Adjustment to graphical model

eteq:
A relatively minor tweak to the graphical model.

I could be wrong in my interpretation of the model here, but I would have said the artificial start tests are not parameters in the model per se, but rather the *results* of the artificial star tests are inputs into the observation model, and the combination of the fluxes and the AST results *is* the observation model.  So then the likelihood is only a result of the observation model.

I also made the change that the ASTs don't depend on the observed fluxes because I would have said that ASTs are whatever they are regardless of the assumption for a *particular star*.  But I could be missing some subtlety so could back that out.

One step further might be to drop the "observation model" entirely as it's sort of implicit that that's where the observed fluxes come from.  But I think even if it's slightly repetitive it's nice that it's clear here.

### DIFF
--- a/beast/plotting/plot_graphic_model.py
+++ b/beast/plotting/plot_graphic_model.py
@@ -69,7 +69,7 @@ def plot_graphic_model(gtype="text", savefig="png"):
         "GC": ("Instrinsic\nDust", "ID"),
         "Obs": ("Observation Model", "<&mu;<sub>i</sub>(F<sup>Mod</sup><sub>i</sub>), &sigma;<sub>i</sub>(F<sup>Mod</sup><sub>i</sub>)>"),
         "Like": ("Observed Band Fluxes", "Observed Band Fluxes"),
-        "AST": ("Artifical Star Test Parameters", "b(m), &sigma;(m)"),
+        "AST": ("Artifical Star Test Parameters", "b(m<sub>i</sub>), &sigma;(m<sub>i</sub>)"),
     }
 
     edges = {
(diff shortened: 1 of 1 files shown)


karllark:
I don't feel like this is correct.  The inputs for the ASTs are from the BEAST physics model.  The results of the ASTs are the bias (mu) and sigma terms that are a function of Fmod.  What we fit is Fmod + mu(Fmod) with sigma(Fmod) as the uncertainties.  Hence, I found having the connection between Fmod to the ASTs was a nice representation that the ASTs are based on Fmod (and the photometry software).

karllark:
beast-graphic

eteq:
Aha, I understand a little bit better now what the variables here are. So that helps me clarify a bit more what I was thinking here: the "AST" node I meant as the *parameters* in mu and sigma (so the i-indexing was wrong in my diagram above!). I'm not clear exactly how the BEAST AST model is implemented, but the example in my head from other AST-based approaches is, say, the polynomial coefficients of a polynomial fit to mu or sigma. So the parameters in that node are the *parameters* of mu and sigma, as opposed to the evaluation of them on the *actual* data. 

With that in mind maybe it's clearer why I don't see why the AST can depend on the physics model - I would say the AST's parameters should be set at least conceptually based on global observational parameters, *not* any specific star.  Does that make sense or am I missing something fundamental here?

And I pushed some commits up to update the obvious bug above and now it looks like this:
image

karllark:
> With that in mind maybe it's clearer why I don't see why the AST can depend on the physics model - I would say the AST's parameters should be set at least conceptually based on global observational parameters, _not_ any specific star. Does that make sense or am I missing something fundamental here?

The observation model is based on the physical model.  Since the observation model *is* a model, it really is applied to the physics model.  This removed the bias of the observation model being based on the data - which has noise the the measurement.  The math works out better.  See the BEAST paper.

In addition, since the photometry is done simultaneously on all the bands (both detection of sources and measurement of the photometry), the observation model formally is a function of both the brightness of a source *and* the spectral shape.  We use the physicsmodel to limit the nD space we have to generate ASTs for - much smaller than a full nD box.

karllark:
Hence the parameters for mu and sigma *are* F^mod_i.

karllark:
And we are not fitting the functions of mu and sigma with analytic functions (maybe we should).  We just are computing them in bins of F^mod_i.

karllark:
Going to close this PR as it has gone dormant.  Please reopen if desired.

PR #636, opened 2020-10-02
removing NaNs from ast results

karllark:
There shouldn't be NaNs in the AST results, but sometimes there are.  Change so that these are filtered out in the toothpick obsmodel construction.

Also fixed a time deprecation.   We were using a very deprecated (since python 3.3) time function.  Found as I'm now using python 3.8.


cmurray-astro (commented):
(see comment above) maybe could include some explanation for why there are NaNs

cmurray-astro (approved):

PR #637, opened 2020-10-05
Edits to supplementing BEAST-optimized ASTs

cmurray-astro:
I made some edits to augment the existing code for supplementing ASTs to the nominal BEAST-optimized set (for MATCH-based applications). This includes:
- adding a column to the output table from pick_models_toothpick_style to record the index of the SEDs used (so they're not duplicated in the supplements)
- adding the ability to impose color cuts in addition to magnitude cuts for the randomly-sampled ASTs

Looking forward to review by @galaxyumi ! :D

### DIFF
--- a/beast/observationmodel/ast/make_ast_input_list.py
+++ b/beast/observationmodel/ast/make_ast_input_list.py
@@ -215,7 +215,14 @@ def pick_models_toothpick_style(
             print(bin_count)
 
     # Gather the selected model seds in a table
-    sedsMags = Table(sedsMags[chosen_idxs, :], names=filters)
+    sedsMags_table = {}
+    for k in range(len(filters)):
+        sedsMags_table[filters[k]] = sedsMags[chosen_idxs, k]
+    sedsMags_table["sedgrid_indx"] = chosen_idxs
+
+    sedsMags = Table(sedsMags_table)
+
+    print(sedsMags)
 
     if outfile is not None:
         ascii.write(
@@ -375,6 +382,7 @@ def supplement_ast(
     outASTfile=None,
     outASTfile_params=None,
     mag_cuts=None,
+    color_cuts=None,
 ):
     """
     Creates an additional fake star catalog from a BEAST model grid
@@ -409,6 +417,9 @@ def supplement_ast(
     mag_cut: dictionary (optional, default=None)
         Ditionary of bright and faint magnitude limits for given filters
 
+    color_cut: dictionary (optional, default=None)
+        Ditionary of faint color limits for given filters
+
     Returns
     -------
     sedsMags: astropy Table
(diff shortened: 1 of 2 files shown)


beast/observationmodel/ast/make_ast_input_list.py
```diff
@@ -215,7 +215,12 @@ def pick_models_toothpick_style(
             print(bin_count)
 
     # Gather the selected model seds in a table
-    sedsMags = Table(sedsMags[chosen_idxs, :], names=filters)
+    sedsMags_table = {}
+    for k, filter_name in enumerate(filters):
+        sedsMags_table[filter_name] = sedsMags[chosen_idxs, k]
+    sedsMags_table["sedgrid_indx"] = chosen_idxs
```
galaxyumi, line 221: sedgrid_indx has been saved in the *ASTparams.fits. You don't need to save the sedgrid_indx in an ascii file if it's unnecessary.

beast/observationmodel/ast/make_ast_input_list.py
```diff
@@ -432,34 +441,53 @@ def supplement_ast(
...
             and remove those SEDs from the SED grid".format(
-                existingASTfile))
+                existingASTfile
+            )
+        )
 
-        t = Table.read(existingASTfile, format="fits")
-        sedsMags = np.delete(sedsMags, t["sedgrid_indx"], axis=0)
-        sedsIndx = np.delete(sedsIndx, t["sedgrid_indx"])
+        t = Table.read(existingASTfile, format="ascii")
```
galaxyumi, line 448: If you are going to decide not to save sedgrid_indx in the ascii file, then this part needs to be reverted.

beast/observationmodel/ast/make_ast_input_list.py
```diff
@@ -432,34 +441,53 @@ def supplement_ast(
...
+            cond = np.logical_and(cond, tmp_cond.ravel())
 
         sedsMags = sedsMags[cond, :]
         sedsIndx = sedsIndx[cond]
 
     # Randomly select models
     # Supplementing ASTs does not need to follow
     # the toothpick-way selection
-    chosen_idxs = np.random.choices(np.arange(len(sedsIndx)), k=nAST)
+    chosen_idxs = np.random.choice(np.arange(len(sedsIndx)), nAST)
```
galaxyumi, line 484: Oh! It seems like we can just provide an int for the first parameter, i.e., np.random.choice(len(sedsIndx), nAST).

a : 1-D array-like or int
    If an ndarray, a random sample is generated from its elements.
    If an int, the random sample is generated as if a were np.arange(a)

cmurray-astro:
this is still a WIP! I'm going to add a test.

galaxyumi (commented):
I left just a few comments. Everything else looks good to me!

cmurray-astro:
@karllark this should be good to go! thanks to Yumi's review the only change is to add color cut capabilities to the supplemental ASTs :)

Issue #616, opened 2020-09-08
Systematically check use of np.empty


karllark:
There are places in the code where np.empty is used.  This can cause problems if the uninitialized elements are used.  All these instances should be checked and probably converted to np.zeros or np.full.

petiay:
Working on this now. (Let me know if there is a better way to make a note that I am addressing an issue).

PR #642, opened 2020-10-16
Replaced all occurences of np.empty() with np.zeros()

petiay:
Addressing issue #616  - https://github.com/BEAST-Fitting/beast/issues/616


karllark (approved):
Looks great.  Thanks.

PR #644, opened 2020-10-16
Add more options to use simulate_obs to properly simulate ASTs

galaxyumi:
I added two more optional keyword parameters in 'simulate_obs': complcut for a completeness cut and magcut for a magnitude in the 'compl_filter'. Recommend to use one or the other. If these cuts are supplied, the selected model SEDs will be limited to those with completeness >= complcut or magnitude <= magcut in the 'compl_filter'. These can be particularly useful when simulating additional ASTs.  

I also added 'uniform' weight option for the 'weight_to_use' keyword parameter. This option is also useful when simulating ASTs by avoiding the skewness to lower-mass stars.

(Example) 
beast simulate_obs physicseds.grid.hd5 noisemodel.grid.hd5 output.fits --nsim 100000 --compl_filter F475W --magcut 30 --weight_to_use uniform

### DIFF
--- a/beast/observationmodel/observations.py
+++ b/beast/observationmodel/observations.py
@@ -217,6 +217,7 @@ def gen_SimObs_from_sedgrid(
     sedgrid_noisemodel,
     nsim=100,
     compl_filter="F475W",
+    complcut=None,
     ranseed=None,
     vega_fname=None,
     weight_to_use="weight",
@@ -247,6 +248,10 @@ def gen_SimObs_from_sedgrid(
     compl_filter : str
         Filter to use for completeness (required for toothpick model).
         Set to max to use the max value in all filters.
+
+    complcut : float (defualt=None)
+        completeness cut for only including model seds above the cut
+        raning from 0.0 to 1.0.
 
     ranseed : int
         used to set the seed to make the results reproducable,
@@ -280,8 +285,8 @@ def gen_SimObs_from_sedgrid(
     model_compl = sedgrid_noisemodel["completeness"]
 
     # only use models that have non-zero completeness in all filters
-    #  zero completeness means the observation model is not defined for that filters/flux
-    ast_defined = model_compl > 0
+    # zero completeness means the observation model is not defined for that filters/flux
+    ast_defined = model_compl > 0.0
     sum_ast_defined = np.sum(ast_defined, axis=1)
     goodobsmod = sum_ast_defined >= n_filters
 
@@ -305,6 +310,11 @@ def gen_SimObs_from_sedgrid(
         filter_k = short_filters.index(compl_filter.upper())
         print("Completeness from %s" % sedgrid.filters[filter_k])
         model_compl = model_compl[:, filter_k]
+
+    # if complcut is provided, only use models above that completeness cut 
+    # in addition to the non-zero completeness criterion
+    if complcut is not None:
+        goodobsmod = (goodobsmod) & (model_compl >= complcut)
 
     # initialize the random number generator
     rangen = default_rng(ranseed)
--- a/beast/observationmodel/observations.py
+++ b/beast/observationmodel/observations.py
@@ -218,6 +218,7 @@ def gen_SimObs_from_sedgrid(
     nsim=100,
     compl_filter="F475W",
     complcut=None,
+    magcut=None,
     ranseed=None,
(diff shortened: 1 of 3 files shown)


docs/simulations.rst
```diff
@@ -32,6 +32,14 @@ dividing by the average mass (mass prior model).  This number of stars is
 simulated at each age and then the completeness function is applied to remove
 all the simulated stars that would not be observed.
 
+The `beast simulate_obs` also can be used to just simulate extra ASTs. This is very
+powerful if you need more ASTs for generating MATCH-optimized ASTs in addition to what 
+you already have from the DOLPHOT ASTs. This means that you can avoid extensive/additional
+DOLPHOT ASTs as long as your initial/existing DOLPHOT ASTs properly cover the entire
+CMD space (i.e., meeting the minimum requirement by MATCH should be enough.) by
+providing the information on the flux uncertainties, biases, and the completeness
+for the additionally selected model SEDs.
```
karllark, line 41: I feel a bit more detail and explanation would be good here.  Maybe state that the simulation is *replacing* generating additional input ASTs and running those through DOLPHOT.  Might be useful to state that the initial/existing ASTs generated to be optimized for the toothpick model (BEAST-optimized?), provide the needed coverage.

Also, I think it would be better to be general about the photometry routine as there is not specific dependence on ASTs and the BEAST for DOLPHOT (even if that is really only what we've used).

beast/observationmodel/observations.py
```diff
@@ -248,6 +250,14 @@ def gen_SimObs_from_sedgrid(
         Filter to use for completeness (required for toothpick model).
         Set to max to use the max value in all filters.
 
+    complcut : float (defualt=None)
+        completeness cut for only including model seds above the cut
+        raning from 0.0 to 1.0.
```
karllark, line 255: raning -> ranging
Maybe "where the completeness ranges between 0 and 1."

beast/observationmodel/observations.py
```diff
@@ -371,15 +396,19 @@ def gen_SimObs_from_sedgrid(
...
 
     else:  # total number of stars to simulate set by command line input
-        gridweights = sedgrid[weight_to_use][goodobsmod] * model_compl[goodobsmod]
-        gridweights = gridweights / np.sum(gridweights)
 
-        # sample to get the indexes of the picked models
-        sim_indx = rangen.choice(model_indx[goodobsmod], size=nsim, p=gridweights)
-        print(f"number of simulated stars = {nsim}")
+        if weight_to_use == "uniform":
+            # sample to get the indexes of the picked models
```
karllark, line 401: indexes -> indices

karllark (changes requested):
Great to see the progress on simulating ASTs!  Just a few minor changes requested.

docs/simulations.rst
galaxyumi: I tried to reflect your comments as much as possible. Based on my tests so far, we need BEAST-optimized (toothpick) + MATCH-friendly(random in given magnitude and color ranges) as the existing ASTs from which we construct a noisemodel grid. But the real ASTs do not need to densely cover the CMD; uniform and complete coverage is more important. We can update this documentation later when we have more robust test results.

beast/observationmodel/observations.py
galaxyumi: Fixed.

beast/observationmodel/observations.py
galaxyumi: Fixed.

beast/tools/simulate_obs.py
```diff
@@ -52,20 +54,34 @@ def simulate_obs(
         filter to use for completeness (required for toothpick model)
         set to max to use the max value in all filters
 
+    complcut : float (defualt=None)
+        completeness cut for only including model seds above the cut
+        raning from 0.0 to 1.0.
```
karllark, line 59: raning -> ranging

karllark (changes requested):
Looks great.  Minor typo needs fixing and then this PR is ready for merging!

beast/tools/simulate_obs.py
galaxyumi: Done!

karllark (approved):
Looks great!

PR #645, opened 2020-10-16
add code that plots ast info

christinawlindberg:
Code for viewing the bias, uncertainty, and completeness. You can choose whether you want to interpolate between the flux bins or not. 

image
ASTs with just 2 source density bins but lots of filters

image
ASTs with lots of source density bins but only 2 filters


karllark (changes requested):
I think a legend of some sort would be very useful for understanding these plots.
Also, please fix the LGTM identified error.

karllark:
Also, it would be great to add this script to the `beast/main.py` so that it can be called via `beast plot_ast_diagnostics`.

Now that I think about it, I'm not sure that plot_ast_diagnostics is the best name for this function/script/plot.  Maybe something that includes observation model and source density bins as this is specifically what it is showing.  I'm thinking we might have more "ast" diagnostic plots in the near future and it would be good to be as descriptive as possible.  `plot_toothpick_sourceden`?  Something like that at least.

christinawlindberg:
> I think a legend of some sort would be very useful for understanding these plots.

What sort of legend info do you think would make the plot more clear? I can't think of anything that isn't repeating some information already on the plot.

> Now that I think about it, I'm not sure that plot_ast_diagnostics is the best name for this function/script/plot. Maybe something that includes observation model and source density bins as this is specifically what it is showing. I'm thinking we might have more "ast" diagnostic plots in the near future and it would be good to be as descriptive as possible. plot_toothpick_sourceden? Something like that at least.

I agree. Although the plot can be made regardless of whether there are source density bins or not (I've added an example below). So my inclination would be to call it `plot_toothpick_obsmodel` but that's pretty similar and could get confused with `plot_toothpick_details`. Would it be a bad idea to call `plot_toothpick_details`  `plot_toothpick_completeness` instead?

image

> Also, it would be great to add this script to the beast/main.py so that it can be called via beast plot_ast_diagnostics.

I'll add the line to `main.py` once we've settled on a name.

karllark:
> I agree. Although the plot can be made regardless of whether there are source density bins or not (I've added an example below). So my inclination would be to call it `plot_toothpick_obsmodel` but that's pretty similar and could get confused with `plot_toothpick_details`. Would it be a bad idea to call `plot_toothpick_details` `plot_toothpick_completeness` instead?

I think `plot_toothpick_obsmodel` is a great name.  Go for it.  I like `plot_toothpick_details` as it is about the details of the model including the individual ASTs results (and not just about completeness).

karllark:
> What sort of legend info do you think would make the plot more clear? I can't think of anything that isn't repeating some information already on the plot.

When there are multiple lines in a single plot, a legend should be there so that what the different lines are is explained.

karllark:
What are the dips in the abs bias plots?  Are these because you are plotting abs bias and not just bias?  What does the bias plot look like?  Plotting just the bias could make more sense.

christinawlindberg:
image

I tried taking away the abs() from the bias (ignore the y-label) and it still looks super strange just for thee F336W filter. I'm not sure why this is the case but this plot was made from the new Scylla ASTs.

karllark:
The plots are good.  With the "scale" bar to say what the color means for the source density, the legend is not needed.

I like the plots in true bias (not abs(bias)).   This is showing what is the actual bias and even if it is odd, it is great to see the actual bias.  I don't think we should focus on the oscillations at this point.  I have a feeling that much of this will go away once we account for the not-really-recovered sources at faint levels.

karllark:
Why are the values on the x-axis negative?

karllark:
Any progress on this PR?

christinawlindberg:
Kinda, although I'll probably edit the function again once we start getting the first Scylla field results. :)

1.

karllark:
Can you address the codacy issue?  Then I'll do another review so we can merge it soon.  Better to get it into the BEAST and the update it later I feel.

karllark:
Can you please address the LGTM errors?  Thanks.

karllark (approved):

PR #648, opened 2020-10-16
APE17 updates

### LINKED ISSUE #586: Install/test docs need updating
We have switched to the new APE17 packaging.  This means it is (for example) `tox -e tests -- --remote-data` instead of `python setup.py test --remote-data`.  See https://docs.astropy.org/projects/package-template/en/latest/ape17.html#step-12-setting-up-tox for more details.

Various places in the docs need updating.  How to install.  Developer docs.

petiay:
Closes #586

### DIFF
--- a/docs/beast_development.rst
+++ b/docs/beast_development.rst
@@ -191,13 +191,13 @@ be viewed in a web browser by pointing to files in the `htmlconv` subdirectory
 
   .. code-block:: console
 
-     $ python setup.py test --remote-data --coverage
+     $ tox -e test -- --remote-data --coverage
 
 Make sure the documentation can be created.
 
   .. code-block:: console
 
-     $ python setup.py build_docs
+     $ tox -e build_docs
 
 The resulting HTML files are placed in the `docs/_build/html` subdirectory, and
 can be viewed in a web browser.
--- a/docs/beast_development.rst
+++ b/docs/beast_development.rst
@@ -76,7 +76,7 @@ with:
 
 .. code-block:: console
 
-   $ python setup.py develop
+   $ tox -e develop
 
 
 Adding Branches
--- a/docs/beast_development.rst
+++ b/docs/beast_development.rst
@@ -71,13 +71,6 @@ In order to run a development installation, navigate to the directory in your
 
    $ pip install -e .
 
-Alternatively, you can perform a development install directly through Python
-with:
-
-.. code-block:: console
-
-   $ tox -e develop
-
 
 Adding Branches
 ===============
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -89,7 +89,7 @@ downloading from Github
 
 .. code-block:: console
 
-     $ python setup.py install
+     $ tox -e install
 
 If you are using conda, you may wish to create a conda environment with the
 dependencies before doing the install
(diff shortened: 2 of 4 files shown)


karllark (changes requested):
This looks great.  Could you also remove any use of `python setup.py develop` or `python setup.py install`?  I noticed at least one of these it none of the `python setup.py` commands are the way to do anything in the new package.  Thanks!

docs/install.rst
```diff
@@ -89,7 +89,7 @@ downloading from Github
 
 .. code-block:: console
 
-     $ python setup.py install
+     $ tox -e install
```
karllark, line 92: The command for this should be `pip install .` (last period is important).

docs/install.rst
```diff
@@ -98,14 +98,14 @@ dependencies before doing the install
 
      $ conda env create -n beast --file conda-environment.yml
      $ conda activate beast
-     $ python setup.py install
+     $ tox -e install
```
karllark, line 101: The command for this should be `pip install .` (last period is important).

docs/install.rst
```diff
@@ -98,14 +98,14 @@ dependencies before doing the install
...
 
 If you would like to modify beast, you may want to use links instead of
 installing, which is best done by replacing the last line with
 
 .. code-block:: console
 
-     $ python setup.py develop
+     $ tox -e develop
```
karllark, line 108: The command for this is `pip install -e .` (the last period is important).

docs/install.rst
```diff
@@ -157,7 +157,7 @@ you can make separate environments for development and production modes.
 
   .. code-block:: console
 
-    $ python setup.py develop
+    $ tox -e develop
```
karllark, line 160: The command for this is `pip install -e .` (the last period is important).

karllark (changes requested):
Thanks!  A few more updates needed.  `tox` does not handle installing, that is now pip.  Thanks again for this PR - clearly we had lots of outdated info.

docs/beast_development.rst
```diff
@@ -76,7 +76,7 @@ with:
 
 .. code-block:: console
 
-   $ python setup.py develop
+   $ tox -e develop
```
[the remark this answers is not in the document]
petiay: Oops, will correct this. It seems that for this particular block I can just remove this line since it now doubles with the code just above on line 72: $ `pip install -e .`

karllark:
There are errors on the doc_test and link_test.  The look simple to fix.  Can you take a look and fix them?

docs/install.rst
```diff
@@ -39,6 +39,14 @@ One easy way to obtain the above is through the AstroConda Python stack:
 
     $ conda create -n astroconda stsci
 
+- Ensure the minimum Python requirements above are met. If needed, specify the 
+minimum version required
```
karllark, line 43: These two lines need to be one - i.e., remove the line break between the two.

karllark (changes requested):
Almost there!

docs/install.rst
petiay: Alright, I def. made this more complicated than it needed to be... :| Now the LGTM analysis: C/C++ check is failing.

karllark: You can ignore that.  It seems to occasionally barf.  I just do so.

PR #652, opened 2020-10-30
WIP: new plot for checking/investigating input ASTs

cmurray-astro:
I wrote a new routine to plot some sanity-check level diagnostics for the input ASTs, ideally to be used (along with plot_ast_histogram) to make sure that the input ASTs look good before shipping them off to be processed. So far, this includes plotting their spatial distribution against the source density image (left, below) and, for all filters in the set, plotting CMDs as a function of source density (right, below; currently done in contours to simplify the plot, and entirely-overlapping contours is what we're looking for... i.e. all ASTs are replicated in the same way across all source density bins). 

An example of what this looks like right now is below.

My questions are: is this useful? If so, what else should be included? Or how should the formatting be improved? Should I split this into individual plots (my original thought was to make a one-stop-shop for sanity checks).
15891_SMC-3956ne-9632_inputAST_diagnostic

### DIFF
--- a/beast/plotting/plot_ast_diagnostics.py
+++ b/beast/plotting/plot_ast_diagnostics.py
@@ -1,0 +1,264 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib.pyplot import cm
+from astropy.io import fits
+from astropy.table import Table
+from astropy.wcs import WCS
+from scipy import interpolate
+from scipy.stats import gaussian_kde
+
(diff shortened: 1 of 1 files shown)


karllark:
I do think it is useful - thanks for opening this PR.  I think it does require some documentation to explain the plots and what users should look for.  Also, it would be good to get this into the main.py script in the base beast directory so that this script can be called with `beast script_name`.

Issue #653, opened 2020-10-31
changes to pick_positions in make_ast_xy_list


lea-hagen:
In #634, I changed the inputs to `pick_positions` so that the input SED file wouldn't get overwritten (instead, it made a new output file, just as `pick_positions_from_map` does).  It looks like that got reverted in #643, and I can't figure out why - the commit message mentions bug fixes for source density binning that are unrelated.  My guess is that it was an accidental change, but I want to make sure I didn't miss something while I was away!

@cmurray-astro

lea-hagen:
I just noticed that the call to `pick_positions` in `tools/run/make_ast_inputs.py` still has the changes from #634, so definitely an accidental change 😄

cmurray-astro:
Yea I can't tell why that change was in #643! Sorry about that, must have been accidental.

lea-hagen:
No worries!  I'm working on it now - will probably take 3 minutes :)

Issue #613, opened 2020-09-04
ASTs results a little confusing


karllark:
I've been working on enhancing/fixing the BEAST simulations.  As part of this, I've dived into the toothpick noisemodel to fix issues with the completeness calculation.  Nothing that would impact the BEAST fitting, but things that do impact the BEAST simulations and the MegaBEAST.

I've made a new plot that illustrates all the individual AST results and the binned toothpick model.  This plot is for the AST results for the METAL LMC 14675 (AST results in 14675_LMC-13361nw-11112.gst.fake.fits).  This plot is created using the command `beast plot_toothpick_details` that will be available once PR #605 is done and merged.  Two things:

1) The ASTs for this field do not seem to be deep enough in the NIR (and maybe optical)?  I'm wondering if these ASTs were done while we were still trimming the AST inputs by the observed data range.  Will need to test this again with new AST results (maybe from the soon-to-be-done Scylla test field).

2) There are a number of bright ASTs that are recovered very faint.  This is the origin of the ASTs that are shown very near 1 in the left plots.  The left plots show the results *normalized* by the input fluxes - hence recovering a bright stars very faint results in a value of 1 in these plots.

![14675_LMC-13361nw-11112 gst fake_plot](https://user-images.githubusercontent.com/1554536/92262362-fd4fc000-eea8-11ea-9d8b-4d390893e8a4.png)

A few people that may be interested: @benw1 , @galaxyumi , @cmurray-astro , @christinawlindberg , @lea-hagen

karllark:
I think this has been answered as due to sources at the edges.  But I'm not sure this has been tested.  @cmurray-astro, @galaxyumi, @christinawlindberg, @lea-hagen or anyone else - have time to test this?

cmurray-astro:
I'll have a look at this sometime in the next few days :)

cmurray-astro:
So I checked this plot for the AST results for SMC_6 from Scylla (post removing NaNs) and I don't see the same population (bright sources recovered faint as in above). Can check the METAL field if you can share the file?
![15891_SMC-3956ne-9632 gst fake_plot](https://user-images.githubusercontent.com/16566310/95480992-80f24600-095a-11eb-937f-0b3ae15870e8.png)

karllark:
There are many fewer sources for sure.  But I think I see a few of them in the F336W plot.  I will send you the file (it is only 8.3 Mb) of the METAL ASTs I'm using.

cmurray-astro:
[...]

PR #654, opened 2020-10-31
Add optional erosion to AST placement region

lea-hagen:
This adds an additional option for placing ASTs in the image.  @cmurray-astro found in #613 that many of the invalid ASTs were due to placement along the edge of the image, so I've added the optional `ast_erode_selection_region` setting to shrink the region where ASTs are allowed to be placed.  The default is currently 0.5" (~10 pixels), and it can also be `None` for no region shrinking.  It gets applied in two ways:
* The AST code always creates a convex hull around the photometry catalog and only places ASTs inside of it.  It will erode that convex hull by 0.5" (i.e., remove the outer 0.5" of the region).
* If `ast_coord_boundary` is set, it will erode that by 0.5".  Since this boundary is generally the very outer extremity of the photometry catalog, it ensures no ASTs will be placed within 0.5" of the edge.

I've updated the docs, but still need to add a test.  Also, the reason it can't automatically merge is addressed in #653.

Addresses #613 (possibly closes?)

### DIFF
--- a/beast/observationmodel/ast/make_ast_xy_list.py
+++ b/beast/observationmodel/ast/make_ast_xy_list.py
@@ -211,7 +211,7 @@ def pick_positions_from_map(
         if radec_pos:
             coord_boundary_radec = erode_path(
                 Path(np.array([set_coord_boundary[0], set_coord_boundary[1]]).T),
-                erode_deg
+                erode_deg,
             )
 
     # if region_from_filters is set, define an additional boundary for ASTs
--- a/beast/observationmodel/ast/make_ast_xy_list.py
+++ b/beast/observationmodel/ast/make_ast_xy_list.py
@@ -109,9 +109,10 @@ def pick_positions_from_map(
 
     erode_boundary : None, or float (default=None)
         If provided, this number of arcseconds will be eroded from the region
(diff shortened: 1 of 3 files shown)


lea-hagen:
I realized there aren't any tests for ASTs yet (unless I'm missing something).  I'll wait on testing the new erosion option until https://github.com/BEAST-Fitting/beast-examples/pull/16 is done, so the code to make the expected output can be incorporated into that.  Also still need to add random seed keywords to all of the AST-generating code, which seems outside the scope of this PR.

All that to say, this is ready for review.

cmurray-astro (approved):
LGTM! Should fix the problem seen in #613, except I guess picking the appropriate value to "erode" is TBD?

lea-hagen:
Thanks for reviewing!  I'm basing the 0.5" on what @benw1 had said before for PHAT, so maybe he can weigh in on the default?

karllark:
Is this PR ready for merging?  Not sure if it is waiting for an updated number or if that will be a separate issue/PR?

lea-hagen:
I suppose we can merge and update the default later if it needs to be changed.

cmurray-astro:
sounds good to me!

PR #655, opened 2020-11-02
Updating sedgrid tests

karllark:
h5py update to v3.0 changes int -> np.int64.  Tests updated to handle this.


karllark:
Very simple PR - merging so that tests work for other PRs.

PR #634, opened 2020-10-02
fixes to beast_info, AST pick_positions, and stats merging

lea-hagen:
While addressing #608, I came across a couple things that needed changing.
* The beast_info asdf file couldn't be updated, because the tree was being read in a `with` block, and apparently memory mapping means its data can't be accessed outside the `with`.  That was solved with a `deepcopy`.  Also, I had to change where the av/rv/fA lists were being made in `model_grid.make_extinguished_sed_grid` so that they were available for beast_info even if the SED grid already existed.
* The metal_small example assigns AST positions with `make_ast_xy_list.pick_positions`, which overwrites the inputted AST SED file.  That was being problematic, so that function now has two filename inputs - one for the input chosen SEDs, and another for the output file with positions.
* Updated the merging (subgrids and SD/background splitting) to deal with stats files that have either 1 or 2 HDUs

Closes #639

PR #656, opened 2020-11-02
Revert accidental changes to pick_positions

### LINKED ISSUE #653: changes to pick_positions in make_ast_xy_list

lea-hagen:
Update to `pick_positions` in `make_ast_xy_list.py` from #634 were accidentally reverted in #643.  This restores the changes.

Closes #653


cmurray-astro (approved):
The changes look good to me!

PR #664, opened 2020-11-21
cleanup/fixes for plot_cmd_with_fits.py

lea-hagen:
Cleaned up the code a bit.  Also fixed a problem with the `savefig` keyword (similar to #603) and changed how argparse does the `log_param` keyword.

### DIFF
--- a/beast/plotting/plot_cmd_with_fits.py
+++ b/beast/plotting/plot_cmd_with_fits.py
@@ -156,12 +156,10 @@ def plot_cmd_with_fits(
         cbar_label = "Log " + param
     cbar.ax.set_ylabel(cbar_label, fontsize=13)  # , rotation=-90, va="bottom")
 
-    # figname
-    basename = data_fits_file.replace(".fits", "_cmd")
-
     # save or show fig
     if show_plot:
         if savefig:
+            basename = beast_stats_file.replace(".fits", f"_cmd_{param}")
             fig.savefig("{}.{}".format(basename, savefig))
         else:
             plt.show()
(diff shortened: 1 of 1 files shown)


karllark:
Was the `savefig` changed?  It seems to not have been.  Does it need to be?

lea-hagen:
Yes, `savefig` is still there, but now it comes from `initialize_parser` in `beast.plotting.beastplotlib`, rather than being defined (incorrectly!) in this code.

beast/plotting/plot_cmd_with_fits.py
```diff
@@ -205,26 +203,22 @@ def plot_cmd_with_fits(
...
-        action="store_true",
-        help="choose whether to take the log of `param` for assigning color",
-    )
-    parser.add_argument(
-        "--savefig",
-        action="store_true",
-        help="choose whether to save the figure, or show it",
+        type=int,
+        default=0,
+        help="If you would like to take the log of `param` for assigning color, set this to 1 (True)",
```
karllark, line 208: Why change this to an int when it is behaving as as a boolean?

karllark: Having the default be False should be action of log_param is not set with store_true, ,so not sure why this change is even needed.

karllark (changes requested):

beast/plotting/plot_cmd_with_fits.py
lea-hagen: There were some other places we'd changed true/false keywords to ints.  But maybe that was where turning on the keyword made the input False, which was simpler to do as an int (and then changed to boolean).  Happy to change this one back.

karllark: Great.  We should use this argparse capability where we can - less code and easier to understand what is happening including the argparse commandline help.

karllark (approved):
Looks good to me.

PR #674, opened 2020-12-04
Fixed a small bug in make_ast_inputs.py when using random_pick method

ogtelford:
This is a one-line bug fix that only affects pick_method='random_pick' in make_ast_inputs.py. obsdata.data no longer has a resolve_alias() method.


karllark (approved):
Looks great!

PR #677, opened 2020-12-04
inverted T axis, added minor ticks

petiay:
Also, removed the transpose of the T vs L plot in the HR diagram. Not sure why the transpose was being plotted since this was reversing the T and L axes yet the axis labels were not reversed (that is, the x-axis was still T and the y-axis was still L).

### DIFF
--- a/beast/plotting/plot_param_err.py
+++ b/beast/plotting/plot_param_err.py
@@ -3,6 +3,8 @@ import matplotlib.pyplot as plt
 from astropy.io import fits
 from matplotlib.colors import LogNorm
 from scipy.stats import binned_statistic_2d as stat2d
+from matplotlib.ticker import (MultipleLocator, FormatStrFormatter,
+                               AutoMinorLocator)
 
 __all__ = ["plot_param_err"]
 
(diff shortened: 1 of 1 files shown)


karllark:
Can you provide the before/after versions of this plot?  Useful to understand the change.

beast/plotting/plot_param_err.py
```diff
@@ -3,6 +3,7 @@
 from astropy.io import fits
 from matplotlib.colors import LogNorm
 from scipy.stats import binned_statistic_2d as stat2d
+from matplotlib.ticker import (MultipleLocator)
```
karllark, line 6: Very minor.  The () are not needed, especially with one import.

karllark (changes requested):

beast/plotting/plot_param_err.py
petiay: Yes! This is leftover from having more ticker imports, and having to remove them to correct code style errors.

petiay:
> Can you provide the before/after versions of this plot? Useful to understand the change.

Here is a before/after. It turns out that plotting the transpose of log(T) vs log(L) with matplotlib's imshow() is needed after all, therefore it has been returned in an updated commit:
Plot 1: Made with the original code
Plot 2: A simple flip of the x-axis to plot an HR diagram

One addition I may make to this code would be to:
A. Either zoom into the HR as in the Rv panel of Plot 2 above OR
B. Choose a plotting scheme which shows all sources since the current scheme does not.
A) would remove some interesting sources, so perhaps it's not the best option.  Feedback is welcome.

Plot 1
param_unc_orig_transposed_Tincr

Plot 2
param_unc_transposed_Tdecr

karllark:
So, plot 2 is what we get with this PR, right?

petiay:
Correct!

karllark (approved):
Looks great.  Thanks!

PR #678, opened 2020-12-04
Remove incorrectly recovered ASTs

karllark:
Remove ASTs that are recovered by chance alignment with a real source.  This removal is done by removing all sources that have recovered fractional fluxes greater than a threshold (default = 0.5, equivalent to 0.75 mag).

Also, added the option to specify the min/max flux range for the toothpick model bins.  Default is to have the min/max values determined by the min/max of the input AST fluxes.

New behavior with removing ASTs (using plot_toothpick_details).
15891_SMC-3956ne-9632 gst fake_plot

Old behavior
15891_SMC-3956ne-9632 gst fake_plot_nocut

### DIFF
--- a/beast/observationmodel/noisemodel/toothpick.py
+++ b/beast/observationmodel/noisemodel/toothpick.py
@@ -1,4 +1,5 @@
 import math
+import copy
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


cmurray-astro (commented):
This looks good to me, Karl!

karllark:
Great.  Merging.

Issue #651, opened 2020-10-23
Option for log binning of source density


cmurray-astro:
Would be great to have the option to bin source densities on a log scale (in addition to linear, which is the current option). Need to test which is better!

cmurray-astro:
working on this for hackday (12/4/20)!

PR #679, opened 2020-12-04
Adding log binning option for source density

cmurray-astro:
Addressing #651, I am working on enabling log binning in source density. This involves including an additional source density parameter in the beast_settings file (will update the beast-examples/ files as well), called "sd_binmode" which can be "linear" or "log". I also added raised errors for the linear case if both the number of bins and the bin width are set, and errors for the log case if the number of bins is not set.

Currently, the way it works is to find the min and max source densities in the field, and then use the sd_Nbins (number of bins) parameter to determine logarithmically-spaced bins. The minimum is the first value in the image greater than zero (as the sd image contains a lot of zeros on the edges which aren't actually from the data). Looking forward to thoughts!

### DIFF
--- a/beast/observationmodel/ast/make_ast_input_list.py
+++ b/beast/observationmodel/ast/make_ast_input_list.py
@@ -211,8 +211,6 @@ def pick_models_toothpick_style(
                     counter * chunksize, successes, successes / counter / chunksize
                 )
             )
-            print("Bin array:")
-            print(bin_count)
 
     # Gather the selected model seds in a table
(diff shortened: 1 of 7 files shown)


karllark (commented):
Looks great.  I feel some documentation would be good to advertise to users that there is the option of linear/log source density spacings.  Would this be possible/"easy"?

karllark (approved):
Great to have docs!

karllark:
Link test failed due to a timeout - checked manually and this link is fine.  Will merge.

PR #683, opened 2020-12-16
Adding custom source density binning option

cmurray-astro:
First crack at adding option of assigning custom source density (or background density) bin edges via the beast_settings file. 

Currently, the way this works is tools/density_map looks for all of the source density binning parameters (sd_binmode (linear or log), sd_Nbins, sd_binwidth and sd_custom (the list of bin edges)). If custom_bins are set via sd_custom, all other source density binning parameters are superseded. However, if not custom, the code still expects all other binning parameters to have been set (i.e., any not in use should be set to None rather than omitted completely). I can update this to check if they're set or not, but figured we'd want to keep the parameters around for completeness? Wondering what the team's thoughts are!

Also, would be great if @christinawlindberg could review this PR!

### DIFF
--- a/beast/observationmodel/ast/make_ast_xy_list.py
+++ b/beast/observationmodel/ast/make_ast_xy_list.py
@@ -67,8 +67,8 @@ def pick_positions_from_map(
 
     N_bins: int
         The number of bins for the range of background density values.
-        The bins will be picked on a linear grid, ranging from the
-        minimum to the maximum value of the map. Then, each tile will be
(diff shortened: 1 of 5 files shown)


christinawlindberg (commented):
Looks good to me.

christinawlindberg:
@karllark Would it be possible to get this merged so I can begin testing for Scylla?

karllark:
Checked are failing due to unrelated download issues.  Will merge.

PR #685, opened 2020-12-18
Add info about using [options.package_data] in setup.cfg

lea-hagen:
If non-standard files (as in, not `.py`) are added to the beast, they also need to be listed in setup.cfg.  Since @karllark has needed to explain that to me on multiple occasions (thanks Karl!), I decided to add it to the development docs.

### DIFF
--- a/docs/beast_development.rst
+++ b/docs/beast_development.rst
@@ -123,6 +123,11 @@ and `beast-dev1`).
 
      $ git add -A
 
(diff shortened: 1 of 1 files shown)


lea-hagen:
Also, @karllark, you're welcome to re-word if what I wrote could be made more accurate or clear.

karllark (approved):
Looks great!

PR #687, opened 2020-12-29
partial merge to check results before subgrid fitting is complete

lea-hagen:
I wrote this a while back and just realized I never did a PR.  I found that I wanted to look at fits for some of the stars in a field, but since the fitting (with subgrids) was still in progress, I would have to wait a while for it to finish before I could run the merging and see the best fits.  This adds the capability to do a partial merge: it looks through the existing stats/pdf1d files (lnP not currently implemented), figures out which stars have been finished in ALL subgrids, and just merges those stars.  It adds a `partial` label to the output files so it's clear that the files don't have the final results.


beast/tools/run/merge_files.py
```diff
@@ -164,13 +176,20 @@ def merge_files(beast_settings_info, use_sd=True, nsubs=1):
...
+        help="set to True (1) if fitting is partially complete and you want to merge stars that are done (relevant to subgrids only)",
+    )
 
     args = parser.parse_args()
 
     merge_files(
         beast_settings_info=args.beast_settings_file,
         use_sd=bool(args.use_sd),
         nsubs=args.nsubs,
+        partial=bool(args.partial),
```
karllark, line 192: If a boolean is needed, why not just set this in the argparser.add_argument setup?  Wouldn't the following make sense?

`parser.add_argument("--partial", action="store_true", help="text")`

This should do the same and avoid the need for `bool` when passing to merge_files.

docs/workflow.rst
```diff
@@ -404,6 +404,15 @@ been implemented.
...
      $ python -m beast.tools.run.merge_files beast_settings.txt --use_sd 1
 
+If fitting using subgrids, it's possible to do a partial merge.  This is useful
+if you'd like to look at fitting results before all stars have finished fitting
+for all subgrids.  This find stars that have fits in all subgrids and merges
+just those.  Partial merge is only currently implemented for stats and 1D PDFs.
+
+  .. code-block:: console
+
+     $ python -m beast.tools.run.merge_files beast_settings.txt --nsubs 5 --partial 1
```
karllark, line 414: If the previous comment works, then the `1` can be removed from the call.

karllark (changes requested):
Looks good other than a minor comment about argparse.

beast/tools/run/merge_files.py
lea-hagen: That works too!  I was following the pattern from `--use_sd`, but that one is True by default, so turning it off requires using ints.

karllark (approved):
Looks good.

PR #691, opened 2021-01-07
allow best legend location; fix matplotlib 3.3 rqmt

petiay:
An explicitly-specified legend location in the lower right of the individual star fit plot sometimes overlapped the legend with a fit line. `matplotlib.legend` can be asked to choose the best location for minimum overlap. A future update can include a legend location option.

The change in the keyword for symlog was based on a MatplotlibDeprecationWarning received just today:
`MatplotlibDeprecationWarning: The 'linthreshy' parameter of __init__() has been renamed 'linthresh' since Matplotlib 3.3; support for the old name will be dropped two minor releases later.`
The latest Matplotlib minor release - [3.3.3](https://pypi.org/project/matplotlib/#history) - took place recently on Nov. 11, 2020, and I believe this is why we may not have run into this issue yet. (I ran into it when testing the legend location option change).

### DIFF
--- a/beast/plotting/plot_indiv_fit.py
+++ b/beast/plotting/plot_indiv_fit.py
@@ -263,7 +263,7 @@ def plot_indiv_fit(filebase, starnum=0, savefig=False, plotfig=True):
         waves, mod_flux_nd[:, 1], mod_flux_nd[:, 2], color="y", alpha=0.1
     )
 
-    #can introduce a legend loc option if 'best' produces overlap
+    # can introduce a legend loc option if 'best' produces overlap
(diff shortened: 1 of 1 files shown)


karllark:
Can you add a description for this PR (you can edit this - it is the "first" comment)?   It is good practice to provide a short description in a PR so that it is clear what is being added/fixed.

One question that came to my mind was why the change to the keyword for the symlog?

petiay:
Good point - commented above.

karllark:
Thanks for the description - now all makes sense!

karllark (approved):
Looks great.

PR #692, opened 2021-01-07
Suppl ast filename

galaxyumi:
Current way of setting the output file names for the supplemented input AST list can cause a problem by making both the BEAST optimized input list and the supplemented input list have the same file names unless you run the 'make_ast_inputs' with 'ast_supplement = False' in the beast_seetings.txt and then run again 'make_ast_inputs' with 'ast_supplement = True'. This is inconvenient for users and might cause unnecessary issue for their ASTs. So I fixed that issue. 

I also update the doc string for the correct usage of the 'make_ast_inputs' with the --suppl_seds option.


cmurray-astro (approved):
LGTM! thanks, Yumi! :)

PR #695, opened 2021-02-24
Fix for the case where f_A = 0.0

karllark:
Currently, if f_A = 0 in any physics grid, a divide by zero error occurs.

### DIFF
--- a/beast/physicsmodel/dust/extinction.py
+++ b/beast/physicsmodel/dust/extinction.py
@@ -570,7 +570,10 @@ class Gordon16_RvFALaw(ExtinctionLaw):
         """
 
         Rv_B = self.BLaw.Rv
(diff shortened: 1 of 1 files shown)


petiay:
So you wouldn’t want to return Rv=2.74 if f_A=0?

karllark:
> So you wouldn’t want to return Rv=2.74 if f_A=0?

The A component is for the MW.  So Rv=3.1 seemed to make more sense.  Doesn't really matter as f_A = 0.0, so the A component isn't used.  But it is calculated (just the way the code works) and so having a non-zero value is needed.  Figured the average for the MW made the most sense for the value to use.

petiay:
Ah, yes. Well, it looks good to me!

petiay (approved):

PR #698, opened 2021-02-25
Allow to indicate the ref image extension number where the WCS info is stored

galaxyumi:
I encountered a case where my reference image has the WCS information in the 0th extension header of a FITS file. However, the BEAST currently assumes the WCS info is always in the first extension of a given reference image. I modified it to allow to explicitly indicate the extension number where the WCS information is stored. I introduced a new parameter 'ast_reference_image_hdu_extension', which needs to be in 'beast_settings.txt' file.


karllark (commented):
What happens if this is not set in the settings file?  Might be good to have a default in that case.

galaxyumi:
Okay. I will make sure that it still works even if this extension parameter is not set in the settings file.

karllark (approved):
Awesome - great!

PR #699, opened 2021-03-09
Cleanup toothpick observation model calculation and fix simulate_obs bug

### LINKED ISSUE #697: SFR=0 bins in the age_prior_model
I have a SFH solution for a galaxy, which has zero SFR for some time bins. To simulate multi-band photometry based on this SFH solution, I simply copied all the time bins and corresponding SFRs, including the time bins with a zero SFR, into the age_prior_model parameter in the beast_settings.txt. 

When I executed the following command "beast simulate_obs seds.grid.hd5 noisemodel.grid.hd5 synthetic_phot.fits --compl_filter=F475W --beastinfo_list beast_info.asdf", I got the following error message. The SFR=0 bins might cause zeros (or nan values) in the combined prior and grid weights.

Completeness from HST_ACS_WFC_F475W
/Users/ychoi/Research/beast/beast/observationmodel/observations.py:379: RuntimeWarning: invalid value encountered in true_divide
 curweights /= np.sum(curweights)
Traceback (most recent call last):
 File “/Users/ychoi/anaconda3/bin/beast”, line 33, in <module>
  sys.exit(load_entry_point(‘beast’, ‘console_scripts’, ‘beast’)())
 File “/Users/ychoi/Research/beast/beast/main.py”, line 90, in main
  func(**funcargs)
 File “/Users/ychoi/Research/beast/beast/tools/simulate_obs.py”, line 121, in simulate_obs
  ranseed=ranseed,
 File “/Users/ychoi/Research/beast/beast/observationmodel/observations.py”, line 381, in gen_SimObs_from_sedgrid
  model_indx[gmods], size=nsim_curage, p=curweights
 File “_generator.pyx”, line 644, in numpy.random._generator.Generator.choice
ValueError: probabilities contain NaN

karllark:
The removing of faint input AST sources that were erroneously recovered with real, brighter sources has been updated.  The criteria is now to remove any sources with output/input flux ratios above a threshold (default is 2.0).  Previously, sources were removed based on their absolute flux difference/input flux (default = 0.5).  This was incorrect as it removed all output negative sources, did not remove sources symmetrically, and removed sources that were recovered fainter than the input fluxes.  Used the wrong logic basically. 

The toothpick handling of AST outputs has been cleaned up.  Removed the support for ASTs where the returned measurements are in magnitudes and not fluxes.  This was an old format and is not actually correct and should not be used ever.  So removing it to avoid an future issues.

Fixes a bug in the simulation of observations that was caused by an age having zero SFR.
Closes #697.

Updated plot from `plot_toothpick_details` (using the *st* AST results)

15275_UGCA292 gst fake_plot

### DIFF
--- a/beast/observationmodel/noisemodel/toothpick.py
+++ b/beast/observationmodel/noisemodel/toothpick.py
@@ -73,7 +73,7 @@ class MultiFilterASTs(NoiseModel):
         self.vega_flux = vega_flux
 
     def set_data_mappings(
-        self, in_pair=("in", "in"), out_pair=("out", "vega"), upcase=False
+        self, in_pair=("in", "in"), out_pair=("out", "rate"), upcase=False
     ):
         """
         Specify the mapping directly with the interface to PHAT-like ASTs
@@ -90,22 +90,30 @@ class MultiFilterASTs(NoiseModel):
         for k in self.filters:
             external_in = k.split("_")[-1] + "_" + in_pair[1]
             external_out = k.split("_")[-1] + "_" + out_pair[1]
+            external_out_flag = k.split("_")[-1] + "_flag"
+            external_out_mag = k.split("_")[-1] + "_vega"
             if upcase:
                 external_in = external_in.upper()
                 external_out = external_out.upper()
+                external_out_flag = external_out_flag.upper()
+                external_out_mag = external_out_mag.upper()
             else:
                 external_in = external_in.lower()
                 external_out = external_out.lower()
+                external_out_flag = external_out_flag.lower()
+                external_out_mag = external_out_mag.lower()
             self.filter_aliases[k + "_in"] = external_in
             self.filter_aliases[k + "_out"] = external_out
+            self.filter_aliases[k + "_flag"] = external_out_flag
+            self.filter_aliases[k + "_outmag"] = external_out_mag
 
     def _compute_sigma_bins(
         self,
-        magflux_in,
-        magflux_out,
+        mag_in,
+        flux_out,
+        fluxout_mag=None,
         nbins=30,
         min_per_bin=10,
-        completeness_mag_cut=80,
         name_prefix=None,
         asarray=False,
         compute_stddev=False,
@@ -121,15 +129,14 @@ class MultiFilterASTs(NoiseModel):
 
         Parameters
         ----------
-        magflux_in : ndarray
-             AST input mag or flux
+        mag_in : ndarray
+             AST input mag
 
-        magflux_out : ndarray
-             AST output mag or flux
+        flux_out : ndarray
+             AST output flux
 
-        completeness_mag_cut : float
-            magnitude at which consider a star not recovered
-            set to -1 if the magflux_out is in fluxes (not magnitudes)
+        fluxout_mag : ndarray
+            AST output mag, if present used to remove sources with mag > 90
 
         nbins : int, optional
             Number of logrithmically spaced bins between the min/max values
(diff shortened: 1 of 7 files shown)


galaxyumi:
All the fixes look good to me. One thing it seems wrong is the 'absdiff' calculation in the 'toothpick.py' code. I think we wanted to exclude ASTs when abs(mag_in - mag_out) > 0.75, corresponding to 0.5 < flux_out/flux_in < 2. Current way of the 'absdiff' calculation only covers 0.5 < flux_out/flux_in < 1.5.

karllark:
> All the fixes look good to me. One thing it seems wrong is the 'absdiff' calculation in the 'toothpick.py' code. I think we wanted to exclude ASTs when abs(mag_in - mag_out) > 0.75, corresponding to 0.5 < flux_out/flux_in < 2. Current way of the 'absdiff' calculation only covers 0.5 < flux_out/flux_in < 1.5.

Can you be more specific here?  The code in question I believe is:
```
        if ast_nonrecovered_frac is not None:
            (indxs,) = np.where(flux_out != 0.0)
            absdiff = np.absolute((flux_in[indxs] - flux_out[indxs]) / flux_in[indxs])
            (indxs2,) = np.where(absdiff > ast_nonrecovered_frac)
            flux_out[indxs[indxs2]] = 0.0
```
And is cast is fluxes as magnitudes are not defined for negative fluxes.

galaxyumi:
The current way can't take negative flux into account neither. For example, let's say we inject a star with flux_in=1E-20, and it is recovered with flux_out=-0.5E-20. Then the |flux_out-flux_in|/flux_in is 1.5, which is out of the criteria. Actually, any negative flux_out values won't be make the current criteria because this |flux_out-flux_in|/flux_in ratio will be always greater than 1 for the negative flux_out . Does it make sense?

karllark:
> The current way can't take negative flux into account neither. For example, let's say we inject a star with flux_in=1E-20, and it is recovered with flux_out=-0.5E-20. Then the |flux_out-flux_in|/flux_in is 1.5, which is out of the criteria. Actually, any negative flux_out values won't make the current criteria because this |flux_out-flux_in|/flux_in ratio will be always greater than 1 for the negative flux_out . Does it make sense?

Not yet.  If the value is |flux_out-flux_in|/flux_in = 1.5 and ast_nonrecovered_frac is set to 0.5, then this AST will not be used.  I do not see where you are getting that 0.5 to 1.5 as being rejected and not just anything above 0.5.

karllark:
BTW, looking at the output of plot_toothpick_details.py shows that ASTs with values of |flux_out-flux_in|/flux_in > 0.5 are being rejected.  Hence, my confusion.

karllark:
@galaxyumi Thanks for the video chat.  Understand the problem now.  As discussed on the call, I'm implemented a fix (revised the description of the PR).

beast/observationmodel/noisemodel/toothpick.py
```diff
@@ -375,10 +345,9 @@ def fit_bins(
 
             d = self._compute_sigma_bins(
                 mag_in,
-                magflux_out,
+                flux_out,
                 nbins=nbins,
-                completeness_mag_cut=completeness_mag_cut,
-                ast_nonrecovered_frac=ast_nonrecovered_frac,
+                ast_nonrecovered_ratio=ast_nonrecovered_ratio,
```
galaxyumi, line 350: should this be included as a parameter in the beast_settings? Where can the users set this ratio?

galaxyumi (changes requested):
I left 3 comments for this PR.

beast/observationmodel/noisemodel/toothpick.py
```diff
@@ -175,46 +169,26 @@ def _compute_sigma_bins(
         # check if any NaNs are present, remove if they are
         # NaNs can be present due to the AST pipeline or in cases where
         # there is missing data (e.g., chip gaps)
-        if np.any(np.isnan(magflux_in)):
-            gvals = np.isfinite(magflux_in) & np.isfinite(magflux_out)
-            magflux_in = magflux_in[gvals]
-            magflux_out = magflux_out[gvals]
+        if np.any(np.isnan(mag_in)):
+            gvals = np.isfinite(mag_in) & np.isfinite(flux_out)
```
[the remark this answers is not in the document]
karllark: They shouldn't be.  But in the past this had been an issue.  So, not sure if I should remove this or leave it "just in case."  Thoughts?

beast/observationmodel/noisemodel/toothpick.py
```diff
@@ -175,46 +169,26 @@ def _compute_sigma_bins(
...
+        # a ratio of output/input fluxes.  This removes sources that are below the
         # the faintest detectable flux that are associated with a real nearby
         # source (random chance that happens depending on the source density)
         # based on input threshold ratio
-        if ast_nonrecovered_frac is not None:
+        if ast_nonrecovered_ratio is not None:
             (indxs,) = np.where(flux_out != 0.0)
-            absdiff = np.absolute((flux_in[indxs] - flux_out[indxs]) / flux_in[indxs])
-            (indxs2,) = np.where(absdiff > ast_nonrecovered_frac)
+            absdiff = flux_out[indxs] / flux_in[indxs]
```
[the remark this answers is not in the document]
karllark: Good point.

beast/observationmodel/noisemodel/toothpick.py
karllark: Not sure.  It is a bit of an ad hock parameter and I'm wondering if we should let it be accessible to the general user.  It is possible to change if if you are an expert user via the code.  So, I'm inclined to leave it as it is right now.

beast/observationmodel/noisemodel/toothpick.py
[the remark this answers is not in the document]
galaxyumi: Oh, okay. If this had been an issue before, it would be better to keep it for "just in case".

beast/observationmodel/noisemodel/toothpick.py
galaxyumi: I actually agree with you. Then another question pops out. Why do we want to consider a case where ast_nonrecovered_ratio is None? Not a big deal though. :)

karllark: That case is already included - check out line 188 in the same file.

karllark:
If you are happy with my latest changes, please approve this PR and then I'll merge it.

galaxyumi (approved):

PR #700, opened 2021-03-09
remove unused Nfilter parameter in toothpick style model selection

galaxyumi:
The 'NFilter' parameter is not used anymore for the toothpick style model selection. I removed that parameter from the relevant functions. Sorry about a typo in my commit message.


karllark:
Can you expand a bit on why this is no longer used?  Mainly curious.

galaxyumi:
That parameter was to introduced for our old way of selecting ASTs using 'pick_models'.  The description for this parameter is as follows:
```
Nfilter: integer
        In how many filters a fake star needs to be brighter than the
        mag_cut value 
```
This parameter has nothing to do with toothpick style selection.

karllark (approved):
Great.  Thanks for the explanation.

PR #702, opened 2021-03-09
Modify plot noisemodel and plot_toothpick_detail

galaxyumi:
I made some modification to these two plotting codes to (hopefully) improve figures.

For 'plot_noisemodel', I replaced the  'color' parameter with 'cmap_name' and removed the 'label' parameter. This enables to automatically cycle colors for source density bins for a given color map (default='jet'), as well as to automatically plot a legend for any given number of source density bins. 

For 'plot_toothpick_detail', I made the figure size flexiable based on a given number of filters.

### DIFF
--- a/beast/plotting/plot_noisemodel.py
+++ b/beast/plotting/plot_noisemodel.py
@@ -13,7 +13,7 @@ def plot_noisemodel(
     noise_file_list,
     plot_file,
     samp=100,
-    color=["black", "red", "gold", "lime", "xkcd:azure"],
+    cmap_name='jet',
     label=None,
 ):
     """
@@ -38,8 +38,8 @@ def plot_noisemodel(
         plotting all of the SED points takes a long time for a viewer to load,
         so set this to plot every Nth point
 
-    color : list of strings (default=['black','red','gold','lime','xkcd:azure'])
-        colors to cycle through when making plots
+    cmap_name : string (default=plt.cm.jet)
+        name of a color map to use
 
     label : list of strings (default=None)
         if set, use these labels in a legend for each item in noise_file_list
@@ -68,6 +68,8 @@ def plot_noisemodel(
     plt.rc("axes", linewidth=2)
     plt.rc("xtick.major", width=2)
     plt.rc("ytick.major", width=2)
+
+    cmap = plt.get_cmap(cmap_name)
 
     # go through noise files
     for n, nfile in enumerate(np.atleast_1d(noise_file_list)):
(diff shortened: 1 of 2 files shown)


beast/plotting/plot_noisemodel.py
```diff
@@ -13,8 +13,7 @@ def plot_noisemodel(
     noise_file_list,
     plot_file,
     samp=100,
-    color=["black", "red", "gold", "lime", "xkcd:azure"],
-    label=None,
+    cmap_name='jet',
```
karllark, line 16: Maybe better to pick a different cmap as the default?  See https://matplotlib.org/stable/tutorials/colors/colormaps.html

karllark (changes requested):
Looks good.  But I am not the expert on plot_noisemodel.   Will ask another for a review of this PR too.

christinawlindberg:
Almost everything looks good to me. The one caveat is that the labels correspond with the order the noise model files are read in, which isn't very informative. E.g. if I gather up the noise model files with glob, the ordering is 0, 1, 10, 11, 12, 2, 3, 4, 5, 6, 7, 8, 9 which then gets labeled as 0, 1, 2, 3, 4, 5... Also, if there are any missing noise files, the labels shift, and it gets kinda confusing. I typically just label them according to the number in the actual file name, 

`num_label = int(nfile.split(".")[-2].split("bin")[-1])`

P.S. One minor thing I noticed is how hard this plot is to create from a command line if there are multiple noise model files. I don't know how often someone wants to run this from the command line though.

christinawlindberg:
@galaxyumi I was only able to make it work with one noise model file, which obviously doesn't convey a lot of information...

beast/plotting/plot_noisemodel.py
galaxyumi: changed it to 'viridis'

galaxyumi:
> Almost everything looks good to me. The one caveat is that the labels correspond with the order the noise model files are read in, which isn't very informative. E.g. if I gather up the noise model files with glob, the ordering is 0, 1, 10, 11, 12, 2, 3, 4, 5, 6, 7, 8, 9 which then gets labeled as 0, 1, 2, 3, 4, 5... Also, if there are any missing noise files, the labels shift, and it gets kinda confusing. I typically just label them according to the number in the actual file name,
> 
> `num_label = int(nfile.split(".")[-2].split("bin")[-1])`
> 


Thanks for pointing this out. I added new lines to sort the input noisemodel files and use the correct SD bin numbers for labeling.

beast/plotting/plot_noisemodel.py
```diff
@@ -190,18 +172,10 @@ def plot_noisemodel(
...
-        help="colors to cycle through when making plots",
-    )
-    parser.add_argument(
-        "--label",
+        "--cmap_name",
         type=str,
-        nargs="+",
-        default=None,
-        help="if set, use these labels in a legend for each item in noise_file_list",
+        default="jet",
```
karllark, line 177: Should be viridis, right?

karllark (changes requested):
Look great.  One small change requested.

christinawlindberg (approved):

beast/plotting/plot_noisemodel.py
galaxyumi: Thanks! Fixed it.

karllark (approved):
Looks great.

Issue #693, opened 2021-01-08
Current way of using a source density map can cause a problem


galaxyumi:
Currently, a source density map is computed at the very beginning of the BEAST runs without any convex hall or/and erosion conditions are applied. While helping @kmcquinn and her student Max Newman, I noticed that the highest source density bin was outside their eroded convex hall area, resulting in that their input ASTs were selected in a lower number of source density bins than they set in the beast_settings.txt. 

Thus, we need to either create a source density map after applying the convex hull or/and erosion conditions or to apply the conditions to the pre-generated source density map before computing the source density bins.

galaxyumi:
I encountered the same issue for one of my galaxies. I set the number of source density to be 8 in the beast_settings.txt file, but the two lowest and two highest density bins were outside the convex hull where I don't have a full filter coverage. Thus the input ASTs were only populated into four (out of 8) source density bins, resulting in a lower resolution in the noise models eventually.

karllark:
As you say, this sounds like the convex hull and/or erosion conditions need to be moved to the initial source density map creation.  Is this straightforward?  Do we use the same code?

galaxyumi:
I don't see any problems with using the same code, at least the core part. I believe it should be straightforward, but might require some minor modifications. Need to find some time to work on this.

karllark:
Great to hear.

cmurray-astro:
I'll work on this today! My thought is that it would be better(?)/easier to go with Yumi's latter suggestion from her first post -- apply the same erosion conditions to the full/original SD map before computing the SD bins rather than use the erosion conditions to generate the SD map (because this erosion business is only used in pick_positions_from_map). What do you think, @galaxyumi?

PR #703, opened 2021-03-10
Adding erosion step to source density map creation

cmurray-astro:
Addressing issue #693, I added an "erode_boundary" parameter to the make_source_dens_map function in create_background_density_map. This (I think) will mean that the source density is computed for the same region that ASTs are placed in... in other words, the RA/Dec grid used for computing source density across each field will be the same (i.e., generated using the full catalog), but the value of the source density in each pixel is based on the sources that overlap with the eroded region. 

Very much looking forward to thoughts from @galaxyumi and @karllark!

### DIFF
--- a/beast/observationmodel/ast/make_ast_xy_list.py
+++ b/beast/observationmodel/ast/make_ast_xy_list.py
@@ -321,7 +321,7 @@ def pick_positions_from_map(
 
                 # - erode_boundary
                 # if you only want to erode the boundary and not impose other
-                # coordinate boundary constraints, still discard SD tiles
+                # coordinate boundary constraints, still discard SD tiles that don't overlap
                 if (set_coord_boundary is None) and (erode_boundary is not None):
                     if catalog_boundary_xy and tile_box_xy:
                         if (
--- a/beast/observationmodel/ast/make_ast_xy_list.py
+++ b/beast/observationmodel/ast/make_ast_xy_list.py
@@ -319,6 +319,27 @@ def pick_positions_from_map(
 
                 # discard tile if there's no overlap with user-imposed regions
 
+                # - erode_boundary
+                # if you only want to erode the boundary and not impose other
+                # coordinate boundary constraints, still discard SD tiles
+                if (set_coord_boundary is None) and (erode_boundary is not None):
+                    if catalog_boundary_xy and tile_box_xy:
+                        if (
+                            Polygon(catalog_boundary_xy.vertices)
+                            .intersection(tile_box_xy)
+                            .area
+                            == 0
+                        ):
+                            keep_tile[j] = False
+                    elif catalog_boundary_radec and tile_box_radec:
+                        if (
+                            Polygon(catalog_boundary_radec.vertices)
+                            .intersection(tile_box_radec)
+                            .area
+                            == 0
+                        ):
+                            keep_tile[j] = False
+
                 # - set_coord_boundary
                 if set_coord_boundary is not None:
                     # coord boundary is input in RA/Dec, and tiles are RA/Dec,
--- a/beast/tools/create_background_density_map.py
+++ b/beast/tools/create_background_density_map.py
@@ -564,8 +564,10 @@ def make_source_dens_map(
             pix_box = geometry.box(i, j, i + 1, j + 1)
             # find fractional overlap area
             frac_area = catalog_boundary.intersection(pix_box).area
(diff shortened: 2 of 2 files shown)


cmurray-astro:
this was admittedly a premature PR. working on bugs!

cmurray-astro:
I changed the strategy so that you can pass an optional parameter to create_background_density_map.main_make_map() called "erode_boundary" (which can be set to the same value as the beast_settings.txt parameter "ast_erode_selection_region", e.g.) which will be used to determine the RA/Dec grid on which the source density map is created (previously, the grid used the min/max values in the catalog).

For Scylla SMC_6 this results in a change in SD maps from the first to the second plots:
Screen Shot 2021-03-10 at 4 27 36 PM
Screen Shot 2021-03-10 at 4 28 01 PM

galaxyumi:
Thanks @cmurray-astro for stepping out for this! 

It is not super clear to me why the higher SD bins still appear on the boundary. Maybe thery are very real this time?

cmurray-astro:
Yes -- the troubling thing I've found is the dynamic range in source density depends strongly on the selected parameters for the map (pixel size (e.g., 5''),  number of pixels, and now erosion boundary)

karllark:
Is there something at the boundary that causes high values - maybe in the computation of the area of the "spatial pixels" at the edges?  Maybe we just should not do any regions that are not fully covered inside the erosion boundary?  Or maybe I don't really understand this issue - if so ignore this comment entirely.

cmurray-astro:
these are good questions! @christinawlindberg has graciously agreed to test this PR and check how the pixel coverages are affected :)

galaxyumi:
@cmurray-astro Yes, the dynamic range in source density also strongly depends on the magnitude range over which the source density is calculated. As long as the **relative** source density profile shape (# of pixels vs. source density) looks more or less the same, the varying **absolute** dynamic range, which depends on many factors, shouldn't be a problem. In other words, we should not rely on the absolute source density values. From conversation with @karllark, he pointed out that we don't know the underlying true source densities anyway without a perfect completeness correction. The key might be not reducing the number of source density bins by too much just based on the dynamic range in SD. For example, for one of the LUVIT galaxies, I had two (actually tried many more) versions of SD maps with different combinations of SD map parameters. I noticed that the dynamic range of one SD map goes from 0 to 15 while that of the other map goes from 0 to 7. I wondered if I want to reduce the number of SD bins for the second case because of a narrow dynamic range, and my answer was no. This was mainly because the way of distribution input ASTs across the image. To better follow the observed stellar density profile, to me it was more reasonable to keep the same number SD bins for those two cases. Does it make sense to you as well?

For the moment, my conclusion  about creating a SD map has been reached to the point where I might want to calculate SD twice. First, create an initial SD map that well reflects the observed 2d stellar density profile just to place input ASTs. Once we compute the completeness from real AST results, we can create the final SD map, which will be used to split catalogs, compute noise models, and fit SEDs. In short, I won't worry too much about getting a precise SD map in the step of generating input ASTs. But, of course, we want to make sure the eroding is happening in the expected way!

christinawlindberg:
I looked at where the SD map tiles/pixels were located for high SD ( > 2.32) and it looks like all the tiles in that SD bin are edge cases. Below, I've created plots for each of the tiles, and it looks like these elevated SD values are just a side effect of the tiny areas. 

If I knew the indices for all of the edge tiles, I would suspect that there are also a lot of tiles with lower than average source density. 

To avoid these cases, we thought it might be best to only compute the source density in tiles with a fractional overlap area > = 10%. @cmurray-astro Has an updated source density plot with this implemented and the SD range is a lot smaller!

sd_plot

cmurray-astro:
yep! excluding pixels in the source density map grid which overlap with the catalog by less than 10% results in this: 
Screen Shot 2021-03-11 at 3 29 09 PM

cmurray-astro:
@galaxyumi I agree with your conclusions about considering SD twice -- first for placing input ASTs, and then for splitting/fitting. Based on Christina's analysis, I also think it makes sense to exclude pixels which have <10% coverage by sources, because our measurement of source density for these is based on very small number statistics. I'm wondering what you both think (@karllark too) -- hard code a limit or add another settable parameter?

karllark:
@christinawlindberg : if you are happy with the revised PR, please formally approve.

christinawlindberg (approved):

Issue #704, opened 2021-03-15
WCS SD Tile transformation incorrect 


christinawlindberg:
It looks like the tile transformation from RA/DEC to x/y isn't correct when figuring out which tiles fall within the boundary region in `make_ast_xy_list.py` around lines 286. I noticed this since there were edge tiles in high source density bins that clearly overlapped the boundary area in RA/DEC,  resulting in 4 non-empty source density map bins, but when the code actually ran, only 3 non-empty map bins were filled. 

Currently, the boundary and tile overlap in `make_ast_xy_list.py` is checked in xy coordinates. If the same check is made with RA/DEC coordinates, then the edge tiles remain and aren't removed, resulting in 4 non-empty map bins.

![sd_radec_v_xy](https://user-images.githubusercontent.com/24800168/111196429-a3386180-8593-11eb-84ba-dbc86a42b6c5.png)

PR #706, opened 2021-03-24
Xy box fix

christinawlindberg:
Fixed xy vs ra/dec priority ordering for checking whether a source density tile is within the boundary area or not. This is a short-term fix for Issue #704. Long term we will need to either figure out if shapely allows diagonal boxes or to just phase out xy functionality completely (although I'm personally not in favor of this plan)


cmurray-astro (approved):
LGTM! definitely open an issue re keeping the xy region capability around long term

PR #709, opened 2021-04-02
Add AST cut_flag to toothpick observation model

karllark:
The toothpick model needs to be input with the full AST list with a designation of which ones were not recovered.  The non-recovered designation is given by the `CUT_FLAG` that is set by external-to-the-BEAST processing.  Usually, this flag is set based on photometry parameters (e.g., sharpness, crowding, etc.).  This information is specifically needed for computation of the completeness.

If the `CUT_FLAG` is not part of the AST output catalog passed to the BEAST, an exception will be raised causing the BEAST to stop.

### DIFF
--- a/beast/observationmodel/noisemodel/toothpick.py
+++ b/beast/observationmodel/noisemodel/toothpick.py
@@ -102,6 +102,7 @@ class MultiFilterASTs(NoiseModel):
         self,
         mag_in,
         flux_out,
+        cut_flag,
         nbins=30,
         min_per_bin=10,
         name_prefix=None,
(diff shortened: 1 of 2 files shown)


cmurray-astro (approved):
this lgtm, Karl! thanks!

karllark:
@cmurray-astro: The tests are failing due to the new format needed for the AST output file.  I can maybe fix this more METAL, but doing the "proper" job would require applying the vgst cuts.  Can you point me to code to do this?  Many thanks in advance.

gausspy:
absolutely! the code to apply the vgst cuts is thanks to Petia in the scylla-data repo (https://github.com/Scylla-HST/scylla-data), scylla_phot_culls.py

is this something I can help do?

cmurray-astro:
whoops sorry the above comment is from me, just didn't realize I was signed into another account!

karllark:
If you wanted to do this for the metal_small ast results in the beast-examples repository, I wouldn't say no!  ;-)

cmurray-astro:
Just put in a PR to beast-examples with a new file!

karllark:
> Just put in a PR to beast-examples with a new file!

You are awesome!

PR #710, opened 2021-04-07
remove sedgrid for fetching filters

christinawlindberg:
A new way to fetch filters based on the ast/gst file contents.

### DIFF
--- a/beast/plotting/plot_toothpick_details.py
+++ b/beast/plotting/plot_toothpick_details.py
@@ -1,13 +1,14 @@
 import matplotlib.pyplot as plt
 
+from astropy.table import Table
+
 from beast.observationmodel.noisemodel import toothpick
-from beast.physicsmodel.grid import SEDGrid
 from beast.plotting.beastplotlib import set_params
 
 __all__ = ["plot_toothpick_details"]
 
 
-def plot_toothpick_details(asts_filename, seds_filename, savefig=False):
+def plot_toothpick_details(asts_filename, savefig=False):
     """
     Plot the details of the toothpick noisemodel creation for each filter.
     These plots show the individual AST results as points as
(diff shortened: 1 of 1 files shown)


beast/plotting/plot_toothpick_details.py
```diff
@@ -123,3 +122,58 @@ def plot_toothpick_details(asts_filename, seds_filename, savefig=False):
...
+    beast_filter_names = [
+        "HST_WFC3_F225W",
+        "HST_WFC3_F275W",
+        "HST_WFC3_F336W",
+        "HST_WFC3_F475W",
+        "HST_WFC3_F814W",
+        "HST_WFC3_F110W",
+        "HST_WFC3_F160W",
+        "HST_WFC3_F657N",
+    ]
```
karllark, line 163: This is not a complete list.  For example, the PHAT/SMIDGE observations would fail here as the optical was taken with ACS not WFC3.   Is there anyway to test if there are other Vega filters in the AST file that are not included in these lists and raise a warning/error?

karllark (changes requested):
Great start.  Please fix the minor codestyle error as well.

christinawlindberg:
All the filters should now be fetched directly from the AST/GST file column names, so that's one less file to keep track of. The `fetch_filters` function looks for all the columns with `"VEGA"` and then sorts them according to wavelength. It then adds the `HST_WFC3` prefix and checks that such a filter actually exists in the BEAST in `_ROOT_/vega.hd5`.

christinawlindberg:
@karllark Were there any other things you wanted me to fix?

beast/plotting/plot_toothpick_details.py
```diff
@@ -1,13 +1,16 @@
 import matplotlib.pyplot as plt
+import tables
```
karllark, line 2: It is critical to use tables?  This package is not the greatest python package and I have been slowly trying to remove it from the BEAST to avoid using it.  I have found I can replace this package with astropy.tables and/or h5py.

beast/plotting/plot_toothpick_details.py
```diff
@@ -123,3 +124,60 @@ def plot_toothpick_details(asts_filename, seds_filename, savefig=False):
...
+
+        if num < 200:
+            filters.append(filters.pop(filters.index('{0}'.format(filters[n - rep]))))
+            rep += 1
+
+    # add camera prefix
+    filters = ["HST_WFC3_" + f for f in filters]
+
+    # check to see that the filters actually exist
+    source = "{0}/vega.hd5".format(__ROOT__)
```
karllark, line 171: Why not directly use the filters file?

__default__ = __ROOT__ + "/filters.hd5"

karllark: See https://beast.readthedocs.io/en/latest/beast_setup.html#beast-filters

karllark (changes requested):
Look good.   A couple of comments.

karllark:
Any time to address my review comments?  Would be good to get this PR in.

PR #712, opened 2021-04-16
Replace 'None' with 'np.nan' for non-matched entries

petiay:
This fixes a bug where if one tried to record a fits table with non-matched entries, since those were recorded as 'None', the following error occurred:
`ValueError: Illegal format `object`.`
Replacing 'None' (type=Object) with 'np.nan' (type=float) allows for the non-matched sources to be recorded without an issue.
(This was not an issue for a returned dictionary, if a table is not recorded.)

### DIFF
--- a/beast/tools/compare_spec_type.py
+++ b/beast/tools/compare_spec_type.py
@@ -181,18 +181,18 @@ def compare_spec_type(
 
         # if there's not a close enough match, put in placeholder values
         else:
-            spec_match["spec_teff"].append(None)
-            spec_match["spec_logg"].append(None)
-            spec_match["phot_cat_ind"].append(None)
-            spec_match["stats_cat_ind"].append(None)
-            spec_match["beast_teff_p50"].append(None)
-            spec_match["beast_teff_p16"].append(None)
-            spec_match["beast_teff_p84"].append(None)
-            spec_match["beast_logg_p50"].append(None)
-            spec_match["beast_logg_p16"].append(None)
-            spec_match["beast_logg_p84"].append(None)
-            spec_match["teff_sigma"].append(None)
-            spec_match["logg_sigma"].append(None)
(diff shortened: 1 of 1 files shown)


petiay:
I don't exactly understand these errors... @karllark, do you?

karllark:
The coverage test errors are due to your changing the output table from having None to np.nan.  There is a test against a cached version of this file.

karllark:
If you are happy with these results, I can merge, then regenerate the cached files based on the new code.  That should remove this error.

karllark (commented):
Is the plot function used in a script anywhere?  Or is it something that would be used in python code or a notebook?  Mainly a question to understand.  This question may be that some documentation should be added as part of this PR.  Possible in the tools section.

karllark:
ping.

petiay:
This took awhile:
@karllark After testing, including with the most recent commit, I am happy with the changes and maybe you can regenerate the cached files.
plot_compare_spec_type.py is now used by compare_spec_type if one wants to plot the results. It can also be used on its own (I can later add the ability to plot directly from the command line, or can do this now if you'd like). Below is an example plot.
As for documentation for these - I think you are suggesting I update the tools_api.rst file in the docs folder. Should I follow the convention there and insert something to the effect of:

```
Comparing spectral types
===================

.. automodapi:: beast.tools.compare_spec_type
```
For some reason I can not test what a change like this may do as I get an error when I convert from .rst to .html (I am using docutils.core to do this).

spec_type_match_plot_example

karllark:
Yes, please add to the docs.  Note that the length of the "===" under the heading has to be the same length as the text above it.  This is likely the origin of your making the docs.  

BTW, I usually generate the docs using `tox -e build_docs` in the base beast repository directory.  Is this how you are making the docs? 

```
Comparing spectral types
========================

.. automodapi:: beast.tools.compare_spec_type
```

PR #713, opened 2021-04-20
Added new optional parameter to write_sbatch_file

cmurray-astro:
New parameter can now be included when writing sbatch files for XSEDE -- the number of nodes to use can now be set (e.g., for the Extreme Memory allocation on Bridges2, the minimum number of nodes to use is 24).


karllark:
Anyone need to review this?  Or is it so specific and "minor" so ok to just merge?

cmurray-astro:
I think it's minor enough to just go in?

karllark (approved):
Looks good.

PR #717, opened 2021-05-26
time.clock() is deprecated, updated to time.time()

christinawlindberg:
Need to update the time function since time.clock() is now deprecated.


karllark (approved):
Looks great.

PR #718, opened 2021-05-27
Minor updates for example docs

karllark:
Adding a bit more text to the example docs to clarify which example should be used.
In addition, a new plot that uses updates to the metal_small data file to provide a more interesting example plot.


petiay:
looks good to me.

PR #720, opened 2021-05-27
Places xy/radec position assignments under the correct if statement

### LINKED ISSUE #711: Should the placement of these xy/radec coord assignment statements be switched?
https://github.com/BEAST-Fitting/beast/blob/f0830b210df971d4017d6e1ab3717c71460eacd5/beast/observationmodel/ast/make_ast_xy_list.py#L175-L183

It seems to me the positions assignments may have been inadvertently inserted under the wrong 'if' statement because presumably:
"if xy_pos and **not** radec_pos..."
then radec_positions need to be assigned. Instead, xy_positions are assigned.
Also:
"if radec_pos and **not** xy_pos..."
then xy_positions need to be assigned, but instead radec_positions are assigned.

Switching the two blocks fixes an eventual error of the type, "UnboundLocalError: local variable 'x_positions' referenced before assignment."

petiay:
Fixes #711.

### DIFF
--- a/beast/observationmodel/ast/make_ast_xy_list.py
+++ b/beast/observationmodel/ast/make_ast_xy_list.py
@@ -171,18 +171,6 @@ def pick_positions_from_map(
             ra_positions = catalog.data["ra"][:]
             dec_positions = catalog.data["dec"][:]
 
-    # # if only one of those exists and there's a refimage, convert to the other
-    # if xy_pos and not radec_pos and refimage:
-    #     radec_pos = True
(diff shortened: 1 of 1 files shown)


karllark:
Ping @galaxyumi and @christinawlindberg.  Would one of  your have time to review this PR this week?  Friday is our next BEAST hack session, so maybe on Friday if not before?

galaxyumi (approved):

karllark:
Thanks @galaxyumi !

PR #721, opened 2021-06-09
Set hdu extention

christinawlindberg:
For XSEDE runs, the lack of an hdu extension being set will throw a warning which halts the process.

This is for the fitting step for the Scylla runs.


karllark:
Can you add a short description above so that it is clear what this PR is for?

christinawlindberg:
I added a short description.

karllark (approved):
Looks good.

PR #727, opened 2021-09-01
Improve grid choice for pdf1d

karllark:
The metallicity grid in the BEAST is usually log spaced.  The 1D pPDF has been generated using a linear spacing between the min/max of the metallicity grid.  This mismatch between the physics grid and 1D pPDF spacing causes the 'Exp' and 'p16, p50, p84' values (all derived from the 1D pPDF) causes these values to not best reflect the actual metallicity information as encoded in the full 7D pPDF.

One solution would to setup the 1D pPDF calculation for metallicity to use a log spacing for the histogram x values.  Quite straightforward as we already do this for the Mass.  

A *better* solution (as discussed below) is to setup the 1D pPDF bins to be the specific physics grid values.  This PR does this in the case where the number of unique values of a parameter is less than the max_nbins set for 1D pPDFs (default is 200).  This is a new capability and will be used for more than just metallicity.  It should give the same results as we have been using linear spacing in general for the physics grid.  But this new code should be more robust and general.

I have not updated any of the 2D pPDFs code at this point.  See issue #729.

As part of this work, a check was added in the 1D pPDF that revealed a bug for the symlog quantities (fluxes w/ biases).  The symlog bug only affected subgrid physicsmodels and was related to not removing the original hard-coded symlog pivot point.  That bug has been fixed with this PR.

Also have updated the plot_indiv_fit code to better visualize the metallicity results.

New plot
beast_metal_small_ifit_starnum_1

Old plot

beast_metal_small_ifit_starnum_1


karllark:
> Just to make sure, will this fix work fine for the linearly spaced metallicity grid case?

Yes, but not ideal.  My thoughts are if one was using a linear metallicity gird case, it would likely be over a narrow metallicity range.   Is this your thoughts as well?  If the linearly spaced metallicity grid is over a wide range, then the log spacing of the 1D pPDF will not be ideal and may suffer from the same kind of issues as we were dealing with before this fix. 

It may be possible to make an adaptive grid instead, but would have to think about this.

At a min, this fix should solve the immediate problem and be at least better than what we had before.

galaxyumi:
> Yes, but not ideal. My thoughts are if one was using a linear metallicity gird case, it would likely be over a narrow metallicity range. Is this your thoughts as well? If the linearly spaced metallicity grid is over a wide range, then the log spacing of the 1D pPDF will not be ideal and may suffer from the same kind of issues as we were dealing with before this fix.

This was my concern. Along with this fix, it might be good to update the relevant doc for the users to guide them to use logs-acing in metallicity to properly evaluate 1D PDF and summary stats. 

> It may be possible to make an adaptive grid instead, but would have to think about this.

I haven't looked at this part of the codes, but why we can't use the physics model grid points? Do we need to redefine each grid points when generating 1D PDFs for parameters?

> At a min, this fix should solve the immediate problem and be at least better than what we had before.
Sounds good!

karllark:
> > It may be possible to make an adaptive grid instead, but would have to think about this.
> 
> I haven't looked at this part of the codes, but why we can't use the physics model grid points? Do we need to redefine each grid points when generating 1D PDFs for parameters?

Ok.  I thought about it.  And it was fairly straightforward to implement.  So, this PR now uses the physics grid points when they are few enough - see the PR description above.

I could really use someone checking that this change works for subgrids.   I believe it should and added specific code to address subgrids, but haven't tested versus are real subgrid run.

PR #735, opened 2021-10-15
Increased the precision of the IAU name generating function

### LINKED ISSUE #716: Duplicate Entries in BEAST
I had been working with some of the BEAST stats files (specifically `b02_sd10-11_sub0_stats.fits` from M31 and I noticed there might be duplicate entries with the exact same RA and DEC but different fields. I'm not sure whether these are distinct stars, or if they were somehow duplicated in the pipeline and fit twice.

duplicates

christinawlindberg:
This change increases the precision of the IAU names generated for each star in the BEAST stats files. The previous level of precision generated duplicate names (see issue #716). 

The RA has a higher precision than DEC based on the official [IAU specifications](http://cdsweb.u-strasbg.fr/Dic/iau-spec.html) which state,

> When sexagesimal coordinates are used, the right ascension portion must have one significant digit more than the declination portion.

Closes #716

### DIFF
--- a/beast/fitting/fit.py
+++ b/beast/fitting/fit.py
@@ -868,11 +868,11 @@ def IAU_names_and_extra_info(obsdata, surveyname="PHAT", extraInfo=False):
                 + c.ra.to_string(
                     unit=ap_units.hourangle,
                     sep="",
-                    precision=2,
+                    precision=4,
                     alwayssign=False,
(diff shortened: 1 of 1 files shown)


karllark:
Failing test makes sense.  Will fix the failing test after the next PR (cached files need to be updated as expected).

PR #741, opened 2021-11-01
fixing docs for flux output units

karllark:
Output fluxes in stats file are in physical units, not normalized units (e.g., flux/flux_vega = input RATE values).


karllark:
Oked on slack.

PR #742, opened 2021-11-03
Change frac_area minimum to 50% instead of 10%

### LINKED ISSUE #740: Edge tiles on SD map don't deserve to exist
Based on recent XSEDE experiences, it seems that there are occasions where high SD bins (>2) are populated with only a handful of stars (<5 sources). This to me indicates that we are still having issues with source density bins being abnormally high near the edge of our fields like we saw in issue #703 and as visualized in the following source density maps.

These high SD bins are likely barely populated but end up generating lots of extra ASTs. for no good reason. I suggest we increase the frac_area coverage to at least 20%, if not more.

15891_SMC-2584ne-14274 vgst_source_den_map_plot
15891_SMC-2750ne-8567 vgst_source_den_map_plot

christinawlindberg:
I'm not sure what the second commit is doing in there but the gist is that 1 digit has changed. 

Closes #740


karllark (approved):
Looks great!

PR #749, opened 2022-01-07
Extend correctly extinction curves below 912A

galaxyumi:
To correctly apply dust extinction below 912A, there is need to merge the observed and theoretical extinction curves because the observed ones are not available (or valid) at these short wavelengths. The necessary lines of code were available, but the functionality was off by default. This PR will make the extended extinction curves available for the Gordon16_RvFALaw() by default. 

To make this PR work properly, we need to add MW_Rv*.txt files back to the beast library package. @karllark

### DIFF
--- a/beast/physicsmodel/dust/extinction.py
+++ b/beast/physicsmodel/dust/extinction.py
@@ -517,7 +517,7 @@ class Gordon16_RvFALaw(ExtinctionLaw):
 
         Rv: float
             R(V) of mixture law (default to 3.1)
-       
+
         Alambda: bool
             if set returns +2.5*1./log(10.)*tau, tau otherwise
 
--- a/beast/physicsmodel/dust/extinction.py
+++ b/beast/physicsmodel/dust/extinction.py
@@ -503,7 +503,7 @@ class Gordon16_RvFALaw(ExtinctionLaw):
             np.min([self.ALaw.x_range[1], self.BLaw.x_range[1]]),
         ]
 
-    def function(self, lamb, Av=1, Rv=3.1, Alambda=True, f_A=0.5, **kwargs):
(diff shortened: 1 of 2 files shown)


karllark:
The number of data files that is required to make this work is quite a few (10s of files).  Each file would need a separate line in the beast/config.py file to have them retrieved using the get_libs.py command.  Would it be possible to combine these files into a single file - like as a table where the columns are the different Rv values?

beast/physicsmodel/dust/extinction.py
```diff
@@ -503,7 +503,7 @@ def __init__(self):
             np.min([self.ALaw.x_range[1], self.BLaw.x_range[1]]),
         ]
 
-    def function(self, lamb, Av=1, Rv=3.1, Alambda=True, f_A=0.5, **kwargs):
+    def function(self, lamb, Av=1, Rv=3.1, Alambda=True, f_A=0.5, draine_extend=True, **kwargs):
```
karllark, line 506: Why did you decide that the Draine extension should be done by default?  I would have expected we would not do this by default to allow for the measured extinction curve to be used unless one would like to do something like get the ionized photons.

galaxyumi: Because no one actually goes below F225W and I am only one currently doing. Setting 'draine extension=True' doesn't affect anything above 1695 AA by using the measured extinction curves down to 1695 AA. If I remember correctly, the measured and Draine extinction curves are not that different down to ~1200 AA. If not, please enlighten me. Any cases, I am fine with adding a new parameter in the beast_settings.txt to make this optional.

karllark: I think we should make this optional.  It is somewhat common to ask for predictions in the GALEX FUV band.  So this would affect that case.

PR #757, opened 2022-02-24
Script for generating main libfile

cmurray-astro:
This is a work in progress! Fun with HDF5 formatting made this take longer than I thought, but figured I'd get what I have up ASAP. @galaxyumi plz comment! So far it only generates filters for WFC3 UVIS and IR, still looking to see if there is an easier way to grab all the filter options from stsynphot...

### DIFF
--- a/beast/tools/make_libfile.py
+++ b/beast/tools/make_libfile.py
@@ -88,7 +88,7 @@ def make_libfile():
     galex = ["fuv", "nuv"]
 
     # Open hd5 file for writing
-    hf = h5py.File(__ROOT__ + "filters_v1.hd5", "w")
+    hf = h5py.File(__ROOT__ + "filters.hd5", "w")
 
     # Create group for nice hierarchical structure
     f = hf.create_group("filters")
--- a/beast/tools/make_libfile.py
+++ b/beast/tools/make_libfile.py
@@ -221,7 +221,7 @@ def make_libfile():
         # populate contents lists with relevant information
         tablenames.append(filter_name)
         observatories.append("HST")
-        instruments.append("WFC")
+        instruments.append("WFPC2")
         names.append(newfilt.name)
         norms.append(newfilt.norm.value)
         cwaves.append(newfilt.cl.value)
--- a/beast/tools/make_libfile.py
+++ b/beast/tools/make_libfile.py
@@ -15,68 +15,26 @@ def make_libfile():
     """
     # wfc3_obsmodes_uvis
     wfc3_uvis = [
-        "f200lp",
         "f218w",
         "f225w",
         "f275w",
-        "f280n",
-        "f300x",
         "f336w",
-        "f343n",
-        "f350lp",
-        "f373n",
         "f390m",
         "f390w",
-        "f395n",
         "f410m",
         "f438w",
         "f467m",
-        "f469n",
         "f475w",
-        "f475x",
-        "f487n",
-        "f502n",
         "f547m",
         "f555w",
-        "f600lp",
         "f606w",
         "f621m",
         "f625w",
-        "f631n",
-        "f645n",
-        "f656n",
-        "f657n",
-        "f658n",
-        "f665n",
-        "f673n",
-        "f680n",
         "f689m",
         "f763m",
         "f775w",
         "f814w",
         "f845m",
-        "f850lp",
-        "f953n",
-        "fq232n",
-        "fq243n",
-        "fq378n",
-        "fq387n",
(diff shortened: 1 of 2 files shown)


cmurray-astro:
aaand tests failing because stsynphot is not found, and I suppose they'll always fail because in order to run stsynphot you need to download ~1 GB of supporting files...

karllark:
> aaand tests failing because stsynphot is not found, and I suppose they'll always fail because in order to run stsynphot you need to download ~1 GB of supporting files...

Tests are failing because stsynphot is not a dependency.  This should just be added to the setup.cfg file.  And will not require the supporting files.  Just that python can load the module.  Can provide more details on the needed update to setup.cfg if it is not clear.

beast/tools/make_libfile.py
```diff
@@ -0,0 +1,229 @@
...
+        pwaves.append(newfilt.lpivot.value)
+        comments.append("")
+
+    # smash the contents arrays together
+    contents = np.array(
+        list(
+            zip(
+                tablenames,
+                observatories,
+                filternames,
```
galaxyumi, line 201: I think this should be instruments

beast/tools/make_libfile.py
```diff
@@ -0,0 +1,229 @@
...
+        # append dataset to the hdf5 filters group
+        f.create_dataset(filter_name, data=arr)
+
+        # generate filter instance to compute relevant info
+        newfilt = phot.Filter(wave, bp_avg, name=filter_name)
+
+        # populate contents lists with relevant information
+        tablenames.append(filter_name)
+        observatories.append("HST")
+        filternames.append(mode.upper)
```
galaxyumi, line 151: I think this should be instruments. Items could be WFC3, ACS, WFPC2, etc.

beast/tools/make_libfile.py
```diff
@@ -0,0 +1,229 @@
...
+    # Open hd5 file for writing
+    hf = h5py.File(__ROOT__ + "filters.hd5", "w")
+
+    # Create group for nice hierarchical structure
+    f = hf.create_group("filters")
+
+    # Define arrays for "contents" / descriptive information
+    tablenames = []
+    observatories = []
+    filternames = []
```
galaxyumi, line 109: filternames -> instruments

galaxyumi:
> This is a work in progress! Fun with HDF5 formatting made this take longer than I thought, but figured I'd get what I have up ASAP. @galaxyumi plz comment! So far it only generates filters for WFC3 UVIS and IR, still looking to see if there is an easier way to grab all the filter options from stsynphot...

I left comment about the "INSTRUMENT" column in the content group. Otherwise, the script looks good to me. You are planning to add ACS and WFPC2 filters?

cmurray-astro:
yes! although I will need help to figure out WFPC2... @karllark mentioned figuring out how to combine the 4 chips? if either of you have insight, please let me know!

cmurray-astro:
Ok I added ACS_WFC, WFPC2 and GALEX filters. Please look through the list and verify that my curating is reasonable? It's very easy to include more! Currently, after discussion with Pey-Lian Lim, the best way to list the filters is via the rather unsatisfying custom list of strings...

galaxyumi:
@cmurray-astro Here is the list of filters used in LUVIT in addition to WFC3.
ACS/WFC filters - F475W, F555W, F606W, and F814W
WFPC2 filters - F439W, F555W, F606W, and F814W

galaxyumi:
> Ok I added ACS_WFC, WFPC2 and GALEX filters. Please look through the list and verify that my curating is reasonable? It's very easy to include more! Currently, after discussion with Pey-Lian Lim, the best way to list the filters is via the rather unsatisfying custom list of strings...

I see all the LUVIT filters, as well as the filters in my own HST programs, in the current list. Thanks Claire! As we briefly talked about the ionizing photon filters, I would like to keep them separate (at least for now). The way loading and applying them are different from other filters in the current BEAST codes because we compute the number of photons, not the fluxes in those bands.

karllark:
Looks like a good list of filters to me.

karllark:
Remember to update the title when this PR is ready for review/merging.

cmurray-astro:
ready for review/merging!

galaxyumi:
> 

Looks good to me!

karllark:
Where can I get the new library file?  Need to replace the existing one.

cmurray-astro:
https://www.dropbox.com/s/arx24r24vn2yfge/filters.hd5?dl=0

PR #761, opened 2022-03-10
Fix verifying beast setting

galaxyumi:
The current limiting range for Z in the verifying beast setting process is only valid for the PARSEC model. The PARSEC webform takes Z (1e-4<=Z<=0.02) whereas the MIST webform takes [Z/H] (-4.0<=[Z/H]<=0.5). So when using the MIST model, having negative Z values in the settings file return an error. This PR fixs that error.


karllark (approved):
Looks good.

Issue #755, opened 2022-02-21
Filter bandpasses need updating


karllark:
Work by @galaxyumi revealed that the latest bandpasses downloaded from STScI (https://www.stsci.edu/hst/instrumentation/wfc3/performance/throughputs) were different than the bandpasses in the BEAST filter library.  The differences were especially large for the medium band WFC3 filters.  The BEAST filter library ingested bandpasses downloaded from an STScI web page that had a huge listing of fits files including multiple versions and this page is not currently able to be found.

Interacting with the WFC3 team, it was clear that the best way to get full end-to-end throughput bandpasses is by using the stsynphot python package.

See https://github.com/karllark/beast-books/blob/check_filters/misc/BEAST%20versus%20STSYNPHOT%20Bandpass%20check.ipynb for a comparison of the BEAST and stsynphot bandpasses for many of the filters often used in the BEAST.

Need to use stsynphot and update all the HST bandpass functions in the BEAST filter library.  For instruments with multiple chips (e.g., ACS, WFC3 UVIS) we should average the bandpasses over the chips.  The code to do this updating should be included in the repository and docs written on how to do this.  In addition, possibly setup tests to alert us when/if the bandpass functions change in the future (this could be challenging as stsynphot requires the download of a database).

List of filters in the BEAST: https://beast.readthedocs.io/en/latest/beast_setup.html#beast-filters
We may want to consider removing filters that we cannot document the origin of the bandpass functions.  Especially if we are not using them.

cmurray-astro:
Adding to this, we need to include documentation explaining the new process of how the filter library is created, and how to add new filters.

mfouesneau:
The WFC3 passbands were from ST calspec data in ~2012. I suppose they updated these but I'm surprised these are so different.
 
How do these compare with the SVO service? http://svo2.cab.inta-csic.es/theory/fps/index.php?mode=browse&gname=HST&asttype= 

if you outsource the passband interface you could also solve #759.

karllark:
> The WFC3 passbands were from ST calspec data in ~2012. I suppose they updated these but I'm surprised these are so different.

The issue is that we seem to have only loaded the filter response function *only*.  This did not include there rest of the optical system, especially the responsivity of the detector that significantly cuts off some of the filter leaks.  This is especially important for the medium band WFC3 IR filters, but not the big leaks for the UV filters.

karllark:
[...]

PR #772, opened 2022-04-28
update docs to reflect udpates in filters.hd5

galaxyumi:
We recently found an issue with the filter curves in the filters.hd5 (issue #755), which only included filter response. Thus we updated filters.hd5 with total throughputs using stsynphot (PR #757). The vega.hd5 file also has been correspondingly updated as of 4/28/2022 (shared the updated file with @karllark). This PR is a pure docs update to reflect our update in filter curves.


karllark (approved):
Looks good to me.

PR #773, opened 2022-06-15
Enable splitting large model grids with subgrids for fitting simulations

cmurray-astro:
This adds a script to allow fitting simulated stellar populations with production-size model grids. This version does this by treating the simulated stars as all belonging to a single source density bin, so that all split files are named with "bin0" by default, and then split by flux to make fitting easier.

### DIFF
--- a/beast/tools/split_simulated_catalog.py
+++ b/beast/tools/split_simulated_catalog.py
@@ -3,12 +3,9 @@
 Split a simulated catalog and a set of AST results by number of sources
 (to improve fitting efficiency for large grids).
 """
-import shutil
 import argparse
 import numpy as np
 from astropy.table import Table
-from beast.tools.density_map import BinnedDensityMap
-
 from beast.tools import beast_settings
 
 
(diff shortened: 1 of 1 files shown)


cmurray-astro:
@christinawlindberg plz review thanks!

beast/tools/split_simulated_catalog.py
```diff
@@ -0,0 +1,172 @@
...
+        if string: file name with beast settings
+        if class: beast.tools.beast_settings.beast_settings instance
+
+    catfile : string
+        name of the photometry catalog file
+
+    astfile : string
+        name of the ast catalog file
+
+    n_per_file : int or None (default=6250)
```
christinawlindberg, line 35: ```default=1000```

beast/tools/split_simulated_catalog.py
```diff
@@ -0,0 +1,172 @@
...
+    print("Splitting catalog")
+    split_simulated_catalog(
+        catfile,
+        n_per_file=n_per_file,
+        min_n_subfile=min_n_subfile,
+        sort_col=sort_col,
+    )
+
+
+def split_simulated_catalog(
```
christinawlindberg, line 61: I'm wondering what the point is of having 'split_main' as a function if all it does is immediately run 'split_simulated_catalog', especially since all the inputs and variable descriptions are the same, making the file 50% longer (~50 lines of code for split_main vs ~170 lines total).

christinawlindberg (approved):

beast/tools/split_simulated_catalog.py
cmurray-astro: yea I set it up this way in case we need to add additional functionality. I don't mind the extra lines if it means more flexibility?

cmurray-astro:
also I'm not sure why the code coverage test is failing?

karllark:
The tests are failing on checking the toothpick model versus the cached version.  This started failing in the weekly tests sometime ago.  Not sure why.  There wasn't a PR that I can see around that time.  So, not related to this PR.

PR #775, opened 2022-10-05
add proper param verification for MIST

galaxyumi:
Although available metallicity range for the MIST models is different from the PARSEC models, the current version of parameter verification is optimized for the PARSEC models. This PR should fix the problem.


karllark:
Can you address the test failures?  They are different from the other to two PRs that look to have been failing one way for sometime (different failures than this PR).

galaxyumi:
@karllark It's done.

karllark (approved):
Looks good to me.

karllark:
Test failures should be fixed with recent separate PR.

PR #777, opened 2022-10-07
Clean up the usage of make_ast_xy_list

galaxyumi:
Although there is an option for placing input ASTs over a region based on a customized filter combination, the make_ast_xy_list.py always place input ASTs in the full filter coverage area. I made the customized filter combination option effective. I also modified a few associated files to make codes cleaner by removing redundant argument/keyword in a function and by removing the part where magnitude information is used. The BEAST uses 'RATE' in the rest of its work flow.

### DIFF
--- a/beast/observationmodel/ast/make_ast_xy_list.py
+++ b/beast/observationmodel/ast/make_ast_xy_list.py
@@ -20,12 +20,11 @@ def pick_positions_from_map(
     N_bins,
     bin_width,
     custom_bins,
-    Npermodel,
+    Nrealize,
     outfile=None,
     refimage=None,
     refimage_hdu=1,
     wcs_origin=1,
-    Nrealize=1,
     set_coord_boundary=None,
     region_from_filters=None,
     erode_boundary=None,
@@ -86,6 +85,14 @@ def pick_positions_from_map(
         Each tile will be put into a bin, so that a set of tiles of the
         map is obtained for each range of source density/background values.
 
+    Nrealize: integer
+        The number of times each model should be repeated for each
+        background regime. This is to sample the variance due to
+        variations within each region, for each individual model.
+
+    outfile: str
+        Path to an output file to store input AST information.
+
     refimage: str
         Path to fits image that is used for the positions. If none is
         given, the ra and dec will be put in the x and y output columns
@@ -99,11 +106,6 @@ def pick_positions_from_map(
         As described in the WCS documentation: "the coordinate in the upper
         left corner of the image. In FITS and Fortran standards, this is 1.
         In Numpy and C standards this is 0."
-
-    Nrealize: integer
-        The number of times each model should be repeated for each
-        background regime. This is to sample the variance due to
-        variations within each region, for each individual model.
 
     set_coord_boundary : None, or list of 2 numpy arrays
         If provided, these RA/Dec coordinates will be used to limit the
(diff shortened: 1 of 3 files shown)


karllark:
The test failures are fixed in a previous PR.

beast/observationmodel/ast/make_ast_xy_list.py
```diff
@@ -20,12 +20,11 @@ def pick_positions_from_map(
     N_bins,
     bin_width,
     custom_bins,
-    Npermodel,
+    Nrealize,
```
karllark, line 23: Why change this from a keyword to a passed argument?  This may break other peoples code.   No idea, but would like to hear more why this change.

beast/tools/cut_catalogs.py
```diff
@@ -158,7 +157,7 @@ def make_cuts(cat_file, partial_overlap=False, flagged=False, flag_filter=None):
 
     # read in the catalog
     cat = Table.read(cat_file)
-    filters = [c[0:-5] for c in cat.colnames if "RATE" in c and "RATERR" not in c]
+    filters = [c[0:-5] for c in cat.colnames if "VEGA" in c]
```
karllark, line 160: Why changing from RATE to VEGA.  This switches between fluxes to magnitudes, right?

karllark (commented):
A few clarifications requested.

beast/observationmodel/ast/make_ast_xy_list.py
galaxyumi: This is actually not a change. It's a cleanup and, in fact, a fix. The function takes both Nrealize and Npermodel, which is the same thing, and their usage is mixed up here and there. To remove confusion, I had to pick one and clean up the other. Do you prefer a keyword over a passed argument? Switching from a passed argument to a keyword is not difficult. This change would not break other people's code b/c (1) no one is repeating at the moment; (2) and if someone has been trying to repeat ASTs, they must have an issue that ASTs are not actually repeated.

beast/tools/cut_catalogs.py
galaxyumi: This line is just for figuring out the base filter name. For example, extracting "F475W" from "F475W_RATE". Thus, this if statement does not need to be any complex like "if "RATE" in c and "RATERR" not in c".

karllark: Ok.   So, it worked before.  And you switched as you prefer to use the VEGA column, because...?

beast/observationmodel/ast/make_ast_xy_list.py
karllark: Ah, that makes sense.  Great.  Thanks for doing this!  I am find with your changes.

beast/tools/cut_catalogs.py
karllark: Just trying to make sure I understand this change.

galaxyumi: Yes, it worked before, but the old if statement requires extra job. What I did is just to simplify the if statement.

galaxyumi: When you do `filters = [c[0:-5] for c in cat.colnames if "RATE" in c]`, it returns:
['F110W',
 'F110W_R',
 'F160W',
 'F160W_R',
 'F275W',
 'F275W_R',
 'F336W',
 'F336W_R',
 'F475W',
 'F475W_R',
 'F814W',
 'F814W_R',
 'F127M',
 'F127M_R',
 'F139M',
 'F139M_R',
 'F153M',
 'F153M_R']
 
 To exclude items with '_R', the old if statement added  this extra condition, `and "RATERR" not in c`. We don't need to do this to figure out the filter list.

karllark: Yes.  Will do so now.

Issue #779, opened 2022-11-02
Issue with placing input ASTs for a case with complex region shape


galaxyumi:
While working on one galaxy with a complicated foot print, I found that the BEAST places input ASTs in the area outside the reference image, which is waste of our resources. This happens when the convex hull technique is not valid. 

I am attaching an example where the convex hull method does not work properly. The left panel shows the observed stellar density map where I want to place input ASTs. The right panel shows the density map of input ASTs generated from the BEAST. As you can see from these plots, the BEAST places input ASTs in the left and right sides of the real source density map where they are in fact empty. 

![convexhull_failing_example](https://user-images.githubusercontent.com/5504376/199567735-69248afa-5fde-4c45-9a5a-d025e19af1e8.png)

We need to replace the convex hull method to another one that can handle any kind of complex boundary shapes.

christinawlindberg:
Just brainstorming about this issue a little bit. I don't think the issue is that the convex hull isn't working properly, it's more the case that what you really need is a concave hull, which would allow for inside angles greater than 180 deg). 
![image](https://github.com/BEAST-Fitting/beast/assets/24800168/4e982c42-8686-4fed-b20a-62e15d78352d)

I managed to find a python code snippet online that uses Delaunay triangulation to create a convex hull around a set of datapoints. Using this method would require manually setting an "alpha" value for how pliable the edges should be, meaning that there should probably be some sort of visual diagnostic to check that the appropriate "alpha" value is being set. 
![image](https://github.com/BEAST-Fitting/beast/assets/24800168/aff3fdd3-9d81-471e-8438-13ab31db0ab9)

christinawlindberg:
#789 should fix this issue.

karllark:
PR #789 was closed with a comment that a new PR would be opened.  This does not seem to have happened.  An update would be great.

PR #789, opened 2023-05-26
Update convex hull code

christinawlindberg:
1. Updated convex hull path with concave hull method to make it possible to fit fields with concave footprints. Regular convex footprints are still fittable with this method.
2. Create a diagnostic plot to visualize the area where ASTs can be placed.
E.g. 
image


christinawlindberg:
Would fix #779

galaxyumi:
I tried this on LUVIT galaxies, which have more complicated footprints from multiple cameras as well as more significant density variation within one HST pointing. Unfortunately this doesn't fix the problem reported in #779. Also, the shapely package version seems to affect its usage. The version @christinawlindberg used was 2.0.1 for her development.

karllark:
@christinawlindberg can you response to @galaxyumi's comment above?

karllark:
Closing.   New PR will be opened base don a branch.

PR #795, opened 2023-12-04
Adding in step and absexp priors of dust and dist

karllark:
The visualizations of the Scylla and SMIDGE results have shown that the existing prior model in BEAST is not sufficient.  Especially when considering the physicsmodel needed for the MegaBEAST (=BEAST prior model).  Specifically, the distribution of dust and stars in distance was not modeled as anything other than flat.  The stars are generally centered in the plane of a galaxy, often modeled as falling off with some exponential scale height.  The is usually seen as concentrated into dust clouds that are located at a specific distance.  

Two new prior models are now included "step" (for Av, Rv, fA) and "absexponential" (for dist) to make the priors more realistic.


karllark:
Codacy analysis failure is for a complexity issue.  This complexity is needed, so will ignore.

PR #798, opened 2024-02-14
updating install docs to make them simpler and faster

karllark:
The install instructions are out-of-date.  Updating them to be simpler as pre-installing dependencies is no longer needed.   Also updating python and beast versions where appropriate.


karllark:
Macos seems to crash randomly.  Ignoring.

PR #799, opened 2024-02-14
adding in ACS SBC, WFC3 and ACS narrow, and JWST filters and enhancing the filter plotting routine

### LINKED ISSUE #786: Filters missing from newest vega.hd5 and filters.hd5 library files
This issue is related to the discussion in #755, [PR#757](https://github.com/BEAST-Fitting/beast/pull/757), #759, and #778.

The newest version of the [BEAST library files](https://beast.readthedocs.io/en/latest/install.html) available either via manual download (Oct 2022) or via `beast get_libfiles` does not seem to contain all the filters in the version of the library files from March 2022.
For example, HST_ACS_WFC_F658N is missing. Generally there are 61 vs 189 filters available in the Oct vs the Mar 2022 files. I can not find a particular discussion to trim the list, so perhaps we just need to update the new list to make sure it includes all filters we need. @cmurray-astro , @galaxyumi , @karllark , comments are welcome.

I see some discussion in [PR#757](https://github.com/BEAST-Fitting/beast/pull/757), but since a version of the library files from just after the discussion (Mar'22) includes the larger filter list, but the new version from Oct'22 is trimmed, then I am a bit confused.

### LINKED ISSUE #796: F170W and F255W not plotting correctly in plot_filters
I am attempting to use the plot_filters module from beast.plotting in a Jupyter notebook to see the filter response curves for some HST WFPC2 filters. When I attempt to include either F170W or F255W, I get the following warning and plot.

```
/astro/users/corbinr1/miniconda3/envs/astroconda/lib/python3.7/site-packages/beast/plotting/plot_filters.py:65: RuntimeWarning: Mean of empty slice
  np.nanmean(f.wavelength[f.transmit > 100.0 * ylim[0]]),
posx and posy should be finite values
posx and posy should be finite values
posx and posy should be finite values
posx and posy should be finite values
```

filter_test

This does not occur if I do not include either of the two problem filters which are unlabeled in the plot above. I have attempted to use plot_filters with filterLib set to None and set to the file found in `beast/libs/filters.hd5`, both produce the result above. The commands I used to generate the plot were

```
from matplotlib import pyplot as plt
from beast.plotting import plot_filters
plot_filters.plot_filters(filter_names=['HST_WFPC2_F170W','HST_WFPC2_F255W','HST_WFPC2_F336W','HST_WFPC2_F439W','HST_WFPC2_F555W','HST_WFPC2_F814W'], filterLib=None, xlim=[1000,20000])
plt.savefig('filter_test.png')
```

### LINKED ISSUE #809: Need to add the option to fit F658N
Needed for observations with narrow-band H-alpha filter.

karllark:
Adding HST HRC FUV filters.
Adding in HST WFC3 narrow band filters.
Adding JWST NIRCam, NIRISS, and MIRI filters.
Added in code to make the vega.hd5 file after making the filters.hd5 file.  (thanks @galaxyumi for the code to start from for this).
Added test to ensure the filters.hd5 and vega.hd5 are consistent (e.g. have the same filters).

Added new doc file with plots of all the filters.  This replaces the text list of filter names.

Filters and vega library files updated in both locations for download (i.e., get_libfiles.py and box location).

Closes #757
Closes #796 
Closes #786
Closes #809


karllark:
Also adding in a couple of WFC3 narrow band filters to support other work.

beast/tools/make_libfile.py
```diff
@@ -339,5 +646,66 @@ def make_libfile():
...
+        dtype=[
+            ("FNAME", "S30"),
+            ("CWAVE", "<f8"),
+            ("LUM", "<f8"),
+            ("MAG", "<f8"),
+        ],
+    )
+
+    vega.create_dataset("sed", data=contents)
+    vega.close()
```
galaxyumi, line 706: Would it be good if we close vega_old and filters tables as well?  Or it's not important.

galaxyumi (approved):
The make_vega_libfile function looks good! Thank you.

beast/tools/make_libfile.py
karllark: Actually, do not need to close those as I changed to using the astropy.table QTable function to read them.  They are not kept open the same way that h5py does.   Might be able to change the vega writing to use the same interface, but likely would need to do more testing.  Trying to move to just using astropy.table functions to avoid such "low" level interface issues.

karllark:
Visually confirmed that the new and old filters are *very* similar.   The differences seen in the tests must be due to very small changes in the filter band profiles.  This was check for HST_ACS_WFC_F475W that is the first failure in the regression tests.

The regression tests files need to be updated to account for these *very* small changes.

PR #805, opened 2024-05-14
updating tox for latest python and github actions

karllark:


karllark:
Coverage test failing due to changes in PR #799.  Will fix this test in that PR.

PR #808, opened 2024-05-23
add f658n narrow band 

petiay:
Adding the option to fit the narrow band H-alpha F658W.


karllark:
Confirmed both ACS and WFC3 H-alpha filters added in PR #799.

PR #816, opened 2024-07-30
typo for doc link in README.rst

eteq:
literally a single character fix


eteq:
Ok right one character was too good to be true.  3 now!

karllark (approved):
Looks great.

PR #819, opened 2024-08-13
Remove simulation code from BEAST

karllark:
Simulation code moved to the MegaBEAST.  

Better suited to be there as that is where ensemble work is done *and* where there is code to change the prior weight (used in the sims) without recomputing the full physicsmodel.


cmurray-astro:
LGTM!

PR #821, opened 2024-08-13
modifications to support megabeast simulations

karllark:
The megabeast simulations are based on recomputing the prior weights using the megabeast ensemble model.  This requires computing the prior weights on the full sedgrid at once.   In the beast, the stellar priors are computed on the spectral grid (not sed grid) and this allows for some simplifications.  Code to address computing the stellar priors on the spectral or sed grid is added.  In addition, the dust priors are moved to a function so that the same function can be used in the beast and megabeast.

As part of this work, a new check is added to determine if there are repeat masses in the single age isochrone.  This is the easiest way to tell if stellar prior and grid weights are being computed on the spec or sed grids (spec has no dust and is where it is done for the BEAST).  Surprisingly, there is at least one case where there is a repeated mass in the spec grid - e.g., directly computed from the stellar isochrones.  Thus, the new code produces slightly different results than the previous code as now the repeated masses get the *correct* grid weight whereas before they were not.

In addition, a bug in `make_ast_input_list.py` was found on line 188 when regenerating the regression test files.  There must have been an upstream behavior that allowed this line to work.  It logically should not have worked.  @drvdputt: any insights on this?  This was your code from a few years back.  No worries if it doesn't spark anything.  Just being comprehensive and tagging you.

### DIFF
--- a/beast/observationmodel/ast/make_ast_input_list.py
+++ b/beast/observationmodel/ast/make_ast_input_list.py
@@ -185,7 +185,7 @@ def pick_models_toothpick_style(
             else:
                 # ... do not include this model again, since we will reject it
                 # anyway.
-                include_mask[idxs == rand_idx] = False
+                include_mask[idxs == rand_idx[r]] = False
 
         # Add the approved models
         chosen_idxs.extend(rand_idx[add_these])
--- a/beast/physicsmodel/creategrid.py
+++ b/beast/physicsmodel/creategrid.py
@@ -14,6 +14,7 @@ possibility to generate grids that cannot fit in memory.
     * likelihood computations need to be updated to allow computations even if
       the full grid does not fit in memory
 """
+
 import numpy as np
 import copy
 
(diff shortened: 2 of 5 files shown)


drvdputt:
About line 188: I would expect the original to throw an exception almost every time, since `idxs` and `rand_idx` should be arrays of different sizes. So `idxs == rand_idx` shouldn't mean anything, and just result in a broadcasting error. But maybe the `else` condition that it is was just not that common, and whatever I'm doing there might just be a case of unnecessary optimization.

If you think about it, masking out at most 100k entries (the chunk size used in the algorithm) out of a grid of several million or billion SEDs, perhaps doesn't matter that much. (Although it might have mattered when I was experimenting with much larger chunk sizes).

karllark:
@drvdputt: thanks for the comment.  This line failed when regenerating the regression test suite of files for caching.  As I stated above, this PR fixes a long-standing bug we never know about.  It is small, but it does change the grid weights.  So, I would have thought this line would have been run many times before now.  Anyway, this is why I was theorizing that maybe something upstream changed and some fall back solution is no longer valid.  Regardless, going to move forward with this PR and this change as you agreed it is the correct fix.

PR #825, opened 2025-02-10
replace http with https

galaxyumi:
Querying PARSEC CMD isochrones stops working. This is a quick fix for it by replacing "http" with "https" in its url.


galaxyumi:
@karllark This is a very simple fix to enable the PARSE CMD retrieval to work again.

karllark (approved):
LGTM!

PR #832, opened 2025-05-16
updating docs based on BEAST HackDay

karllark:


karllark:
Test failure an issue with the parsec website, not this PR.

PR #842, opened 2025-09-26
Added kwarg to convert_hdf5_to_fits to consistently trim too-long column names

Stargrazer82301:
Discovered `convert_hdf5_to_fits` fails if the column names in `.phot.hdf5` file are too long. These column names are typically set by the names of the files from MAST on which the photometry was done.

This fix adds an `autotrim` kwarg to the function (default to False). 

This option, if column names are too long, finds the largest contiguous substring common to the column names that are too long for FITS key names. It then makes a shorter version of those column names. it first attempts to just remove `-` and `_` characters from the column names; and after that, it lops off the beginning characters of the largest contiguous common substring (as these are most likely to be generic fluff). The shortened column name substring is then propagated to all column names that contain the original string.

### DIFF
--- a/beast/tools/convert_hdf5_to_fits.py
+++ b/beast/tools/convert_hdf5_to_fits.py
@@ -155,7 +155,7 @@ def replace_largest_common_substring(set0, set1, set2):
             break
 
     # Raise exception if no fix possible
-    if longest == False:
+    if not longest:
         raise Exception(
             "Some column names are too long to write to FITS table, "
             + "but no shared string found to trim"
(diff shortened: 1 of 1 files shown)


Stargrazer82301:
Doc test fail from ACCESS, as with prior PR, and some ADS timeouts? Anyways, not at our end.

Unsure what the coverage issue is... Is that just because the new kwarg isn't True by default, and isn't in the tests?

karllark (commented):
Please see my comments.  Just trying to get my head around this PR.

karllark:
Ping.

PR #846, opened 2025-11-21
fix Rv_A calculation to proper Rv_B for user-input SMC-like dust model

benjamin-gibson:
In beast/beast/physicsmodel/creategrid.py the function _make_dust_fA_valid_points_generator() had hardcoded the value of Rv_B to 2.74. This value is correct for the G03_SMCBar extinction curve but not for any other. This function now accepts the variable Rv_B from the user-selected extinction law.


karllark (approved):
Looks great!

PR #847, opened 2025-11-21
Ast generation speedup

benjamin-gibson:
There was a specific for loop in pick_models_toothpick_style() that took forever looping through each SED to choose which ones to keep as artificial stars. The new code instead loops through filters and un-full bins and vectorizes over the SEDs. This is significantly faster -- new artificial stars can be generated in seconds.

I also implemented the use of the mag_limits() function for pick_models_toothpick_style(), so now users can set the keywords ast_fluxbin_maglimits and ast_filters_inside_maglimits in their beast_settings.txt file (example below). This allows users to only get back model SEDs that have fluxes within some magnitude range (ast_fluxbin_maglimits) for a critical number of filters (ast_filters_inside_maglimits).

```
Example:
# ast_fluxbin_maglimits : dictionary (optional)
#   If supplied, these magnitude limits will be applied to the SED model grids
#   when selecting additional ASTs. This is a dictionary that includes
#   information for the magnitude cuts as a function of the filters included.
# ast_fluxbin_maglimits = None
ast_fluxbin_maglimits = {"HST_ACS_WFC_F555W": [16,30], "HST_ACS_WFC_F814W": [16,30], "HST_WFC3_F275W": [16,30], "HST_WFC3_F336W": [16,30], "HST_WFC3_F438W": [16,30], "HST_WFC3_F555W": [16,30], "HST_WFC3_F814W": [16,30], "JWST_NIRCAM_F200W": [16,30]}

# ast_filters_inside_maglimits : integer
# Number of filters that must be within the magnitude limits
# for an AST to be included in the list (Default = 3)
ast_filters_inside_maglimits = 5
```

### DIFF
--- a/docs/generating_asts.rst
+++ b/docs/generating_asts.rst
@@ -97,7 +97,7 @@ Parameters used by the flux bin method of selecting SEDs
   Format: {"<filter_name>": [<bright_mag>, <faint_mag>], "<filter_name>": [<bright_mag>, <faint_mag>]}
   Example: {"HST_ACS_F555W": [14, 30], "JWST_NIRCAM_F200W": [13, 28]}
 
-- `ast_filters_inside_maglimits : integer (optional, default = 3)
+- `ast_filters_inside_maglimits` : integer (optional, default = 3)
   Number of filters that must be within ast_fluxbin_maglimits for an SED
   to be eligible for selection as an AST.
 


beast/observationmodel/ast/make_ast_input_list.py
```diff
@@ -126,17 +135,37 @@ def pick_models_toothpick_style(
     modelsedgrid = SEDGrid(sedgrid_fname)
 
     sedsMags = -2.5 * np.log10(modelsedgrid.seds[:] / vega_flux)
-    Nseds = sedsMags.shape[0]
-    Nf = sedsMags.shape[1]
-    idxs = np.arange(Nseds)
 
-    # Check if logL=-9.999 model points sliently sneak through
+    # Check if logL=-9.999 model points saliently sneak through
```
karllark: Should be "silently" I believe.

PR #849, opened 2026-01-16
Fixing errors and warnings from tests

karllark:
Specifically the numpy "Conversion of an array with ndim > 0 to a scalar" error that occurs for the new numpy.

Updated the regression test files to reflect the updates to the isochrones downloaded from the CMD website.

Code to handle the warnings that are not informative.

Updated example SED fit plot in the docs to reflect the updated CMD isochrones.


karllark:
Doctest errors are due to slow web or timeouts due to asking too often during this testing.  Not an issue for this PR.

PR #853, opened 2026-02-27
moving to using dust_extinction models only

### LINKED ISSUE #259: Consider dust_extinction package
We could use the dust_extinction package that has been recently developed and is now an astropy affiliated package.  I was the lead of this development.  This would provide the same plus additional dust extinction models for the BEAST.  In addition the testing and vetting has been more rigorous for this package than the current BEAST dust extinction code.    Some work on the interface would be needed.

https://github.com/karllark/dust_extinction

karllark:
Closes #259


karllark:
Slightly different results are due to dust_extinction likely that dust_extinction has a more correct implementation of the dust extinction models.  This is causing the regression test to fail as expected.  Provides a nice demonstration of why only having one place for the dust extinction models is the way to go.  dust_extinction is much more extensively tested than what we had in the beast from back in the day.

galaxyumi:
Hi Karl, while I haven't closely reviewed this PR, I wanted to share a few initial thoughts. I completely agree with unifying the sources for the dust models and transitioning to a better-tested and well-maintained package! This will make the BEAST lighter and much easier to maintain moving forward. One feature I want to ensure we support is the ability to extend the dust extinction curves down to 54.4 eV (or 22.8 nm), either as an optional configuration from the BEAST side or by default from the `dust_extinction` package.

PR #860, opened 2026-04-28
Updating non-LTE, LTE, and AGB stellar libraries

### LINKED ISSUE #180: Add BOSZ stellar atmosphere models
Updated version of Kurucz models - including good wavelength coverage in the JWST range.
Ideally would replace CK04.  Need to pick a reasonable resolution.  And decide on what to do with the extra grid parameters for carbon and alpha-enhancement abundances.  Maybe these are not critical and we should just pick "standard" values (whatever these are).

https://archive.stsci.edu/prepds/bosz/

### LINKED ISSUE #669: When to use AGB models?
Now that we have the Padova/Colibri models incorporated into the downloadable data #357 , we need some criteria for when to use the models for a given star. At the moment we are not assuming any dust for these models. As has been discussed in #258 we may want to add dust as well as better resolution for thermal pulses.

karllark:
The Tlusty non-LTE models have been updated in 2025 to have an enhanced line list and better coverage in the IR.  This stellar library supersedes the existing tlusty library.

The LTE model grid has been updated in the BOSZ 2024 work.  This updated library supercedes the existing Kuruz library.

The AGB model grid has been updated to have the same updated spectral resolution and wavelength range as the other two grids.   This uses the Aringer 2016 grid.  Much of the code for this was adapted from Martha Boyer's code that she shared with me.  These models do not cover the full wavelength range, so a black body with the stellar Teff was used to extend the coverage.  This was done at the shorter wavelengths below 3600 A and longer wavelengths beyond 25 micron.

A common resolution and wavelength range is required if used of multiple libraries at the same time is desired.  While there is code to handle non-similar wavelength grids, it has failures at the edges at this point.

Code is included that generates the libraries as well as plotting the coverage and boundaries.

Spectra extend from 500 A to 32 micron.  The spectral resolution has been set to 2000.

As part of this work, it was discovered that the original tlusty grid had significantly different logT coverage for different metallicities.  Specifically, at Z = 2e-5, 2e-4, 4e-4, and 6.6e-4 the lowest logT was 4.44.  This causes an issue with the interpolation routines that assume the coverage in logT, logg is the same at all metallicities.  There is no clear fix for the old tlusty grid and it should be no longer be used, the new tlusty2025 grid is better in a number of ways.

The new Tlusty2025 grid shows related issue in that at Z=0.000544 the highest logT values are 4.8 and at Z = 0.000561, the lowest logT values are 4.44.  These two metallicities are close together and code has been added to the Tlusty2025 stellib to set the Z values for these to values to be the average of the two (i.e. 0.000553) so that full coverage is available at this Z value and the interpolation routines will work correctly.

Closes #180.
Closes #669.
Closes #858.


karllark:
Here is an example of the incomplete coverage in the old tlusty grid for some metallicities.
image

karllark:
Here is the full coverage - like it should be for all metallicities.  
image

karllark:
doc_test failure is unrelated to this PR.

karllark:
Coverage of the three grids.
image

karllark:
Coverage of the original Tlusty and Kurucz grids.
image

beast/physicsmodel/stars/stellar_libraries/check_slib_coverage.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # slib = stellib.Tlusty2025()
+    # slib = stellib.BOSZ2024()
+    # slib = stellib.Kurucz()
+    slib = stellib.Aringer2016()
+
+    print(np.unique(slib.Z))
+
+    for cz in np.unique(slib.Z.data):
+        bound = get_stellib_boundaries(slib, dlogT=0.0, dlogg=0.0)
```
galaxyumi: If this line doesn't need to be inside the loop, please move it above the `for` loop.

beast/physicsmodel/stars/stellar_libraries/check_slib_coverage.py
```diff
@@ -0,0 +1,31 @@
...
+        # bbox = slib.get_boundaries(dlogT=0.01, dlogg=0.01)
```
galaxyumi: You can clean up unused lines of code. This comments apply to other commented lines.

beast/physicsmodel/stars/stellar_libraries/check_slib_coverage.py
```diff
@@ -0,0 +1,31 @@
...
+        gvals = slib.Z.data == cz
+        plt.plot(slib.logT[gvals], slib.logg[gvals], "ko")
+        # plt.plot(bound[:, 0], bound[:, 1], "b-", label="new")
+        slib.plot_boundary(dlogT=0.0, dlogg=0.0)
+        # plt.plot(bbox[:, 0], bbox[:, 1], "g-", label="in lib")
+        plt.title(cz)
+        plt.xlabel("log(Teff)")
+        plt.ylabel("log(g)")
+        # plt.legend()
+        plt.show()
```
galaxyumi: Please consider using `plt.savefig()` rather than stopping execution with an interactive `plt.show()` loop.

beast/physicsmodel/stars/stellar_libraries/comp_tlusty25_bosz24.py
```diff
@@ -0,0 +1,68 @@
+import matplotlib.pyplot as plt
+import numpy as np
+import astropy.units as u
+from astropy.io import ascii
+
+from helpers import rebin_spectrum
+
+if __name__ == "__main__":  # pragma: no cover
+
+    solar_z = 0.02
```
galaxyumi, line 10: With new models, do we still stick to this old Zsun value? For example, the PARSEC models use Zsun=0.0152.

beast/physicsmodel/stars/stellar_libraries/comp_tlusty25_bosz24.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # read the kurucz spectrum
+    # wavelength first
+    wave_filename = "/home/kgordon/Python/extstar_data/Models/BOSZ2024/r2000/bosz2024_wave_r2000.txt"
```
galaxyumi, line 14: This line contains a hardcoded absolute path pointing to your specific directory. To ensure portability across different environments and for other users, please replace this with a relative path or leverage the standard beast data directory configuration paths. Also the corresponding documentation should be provided.

beast/physicsmodel/stars/stellar_libraries/comp_tlusty25_bosz24.py
```diff
@@ -0,0 +1,68 @@
...
+        model_filename,
+        format="no_header",
+        # fast_reader={"exponent_style": "D"},
+        names=["SFlux", "SCont"],
+    )
+    mspec_lte["Wave"] = mspec_lte_wave["Wave"] * u.angstrom
+
+    # read the tlusty spectrum
+    model_filename = (
+        "/home/kgordon/Python/extstar_data/Models/Tlusty_2023/z100t15000g400v2.spec.gz"
```
galaxyumi, line 33: Same comment as above.

beast/physicsmodel/stars/stellar_libraries/helpers.py
```diff
@@ -0,0 +1,109 @@
...
+    )
+    full_wave_min = twave[0:-1]
+    full_wave_max = twave[1:]
+    full_wave = 0.5 * (full_wave_min + full_wave_max)
+
+    full_flux = np.zeros((npts))
+    full_npts = np.zeros((npts), dtype=int)
+
+    for k in range(npts):
+        (indxs,) = np.where((wave >= full_wave_min[k]) & (wave < full_wave_max[k]))
```
galaxyumi, line 46: Do the boundaries of new lower-resolution bins align with the pixel edges of the high-resolution input spectrum? I guess not. If an input spans the boundary (e.g., 40% of its width is in bin A and 60% is in bin B), this code assigs the entire flux of that pixel to one bin and zero to the other based on where its center point of edge falls. It might not be a significant fraction of flux, but something to think about.

beast/physicsmodel/stars/stellar_libraries/helpers.py
```diff
@@ -0,0 +1,109 @@
...
+    twave = np.logspace(
+        np.log10(wave_range[0]), np.log10(wave_range[1]), num=npts + 1, endpoint=True
+    )
+    full_wave_min = twave[0:-1]
+    full_wave_max = twave[1:]
+    full_wave = 0.5 * (full_wave_min + full_wave_max)
+
+    full_flux = np.zeros((npts))
+    full_npts = np.zeros((npts), dtype=int)
+
```
galaxyumi, line 44: I tried to come up with a block of code that could replace the rest of this function if this is doing what you intended, which has the following benefit:

> 1. By sampling the cumulative distribution function at exactly `twave`, pixels that are split across a boundary have their integrated energy split perfectly proportionally down to decimals. **No flux is lost or duplicated.**
> 2. By explicitly calculating and carrying in_widths (Δλ_input), it will correctly downgrade models even if the source library switches sampling strategies between different regimes (e.g., if a library samples line cores more densely than the continuum).

```
# Define edge boundaries for the input high-res pixels
# If the input grid contains center coordinates, estimate its bin edges
in_edges = np.zeros(len(wave) + 1)
in_edges[1:-1] = 0.5 * (wave[0:-1] + wave[1:])
in_edges[0] = wave[0] - (wave[1] - wave[0]) * 0.5
in_edges[-1] = wave[-1] + (wave[-1] - wave[-2]) * 0.5

# Calculate total energy (Flux * delta_lambda) per input bin
in_widths = np.diff(in_edges)
energy = flux * in_widths

# Compute the cumulative integrated energy distribution
cum_energy = np.zeros(len(in_edges))
cum_energy[1:] = np.cumsum(energy)

# Interpolate cumulative energy onto the new output grid edges
# Use linear interpolation because cum_energy is stepwise linear (if flux constant in bins)
# or piecewise quadratic (if trapezoidal), but linear on step integrals keeps exact step conservation.
energy_interp = interp1d(
    in_edges, cum_energy, kind="linear", bounds_error=False, fill_value=0.0
)

new_cum_energy = energy_interp(twave)

# Total energy integrated inside each new bin window
new_bin_energy = np.diff(new_cum_energy)

# Convert energy back to flux density (divide by new bin widths)
new_widths = np.diff(twave)
full_flux = new_bin_energy / new_widths

# Track contributing pixels & catch out-of-bounds
full_npts = np.zeros(npts, dtype=int)
for k in range(npts):
    # Kept for compatibility with BEAST's diagnostic tracking downstream
    full_npts[k] = np.sum((wave >= full_wave_min[k]) & (wave < full_wave_max[k]))

# If the output grid expands outside the range of the input library,
# those bins will have accumulated 0 energy and have zero widths or false zeros.
# We flag them as non-finite or zero out-of-bounds where the input didn't reach.
out_of_bounds = (twave[:-1] < in_edges[0]) | (
    twave[1:] > in_edges[-1]
)
full_flux[out_of_bounds] = 0.0
full_npts[out_of_bounds] = 0

return (full_wave, full_flux, full_npts)
```

beast/physicsmodel/stars/stellar_libraries/plot_multiple_slib_coverage.py
```diff
@@ -0,0 +1,29 @@
...
+
+if __name__ == "__main__":  # pragma: no cover
+
+    slibs = [stellib.Tlusty2025(),
+             stellib.BOSZ2024(),
+             stellib.Aringer2016()]
+    labels = ["Tlusty2025",
+              "BOSZ2025",
+              "Aringer2016"]
+    # slibs = [stellib.Tlusty(),
```
galaxyumi: Clean up unused example lists.

galaxyumi (changes requested):
Thanks for this updating! I left some specific comments for the changed files, but I do have a general question. Do we account for alpha-abundance mapping at all?

beast/physicsmodel/stars/stellar_libraries/check_slib_coverage.py
karllark: Good point.  Moved.

beast/physicsmodel/stars/stellar_libraries/check_slib_coverage.py
karllark: Absolutely.  Good point.  Done.

beast/physicsmodel/stars/stellar_libraries/check_slib_coverage.py
karllark: Added option to either save as a png or interactively show.

beast/physicsmodel/stars/stellar_libraries/comp_tlusty25_bosz24.py
karllark: The solar_z value for each model is different.  I've gone back to the papers to figure out the correct value for each grid.  This is given with references in each script that generates the grids.  We want the absolute metallicity, not relative to solar as *each* grid uses a different solar value - as the solar value changes with time.  Exciting. 

I will update this script to compare models from any two stellar libraries using the grids themselves.  This is needed to ensure that the scaling is correct for all the models.  And provides a great way to compare models from two grids for the same stellar parameters.   Will not resolve this comment till the expanded script is ready.

beast/physicsmodel/stars/stellar_libraries/comp_tlusty25_bosz24.py
karllark: See comment above.

beast/physicsmodel/stars/stellar_libraries/comp_tlusty25_bosz24.py
karllark: Ditto.  ;-)

Issue #863, opened 2026-05-20
astropy.units.errors.UnitConversionError with the extinction_extension code.


galaxyumi:
When utilizing `extLaw = extinction.Generalized_RvFALaw(ALaw=extinction.Generalized_DustExt(curve='F19_D03_extension'), BLaw=extinction.Generalized_DustExt(curve='G03_SMCBar_WD01_extension'))`, the script `extinction_extension.py` encounters a `UnitConversionError`, stating that the 'greater' function can only be applied to dimensionless quantities.

PR #864, opened 2026-05-20
Fix mixed unit error between angstroms and wavenumbers in `extinction_extension.py`

galaxyumi:
The bug was caused by two main issues within the `extinction_extension.py`: (1) a mixed usage of angstroms and wavenumbers, and (2) inconsistent input formatting requirements, where certain steps expected raw numerical values while others required values paired with Astropy units. I am opening this PR to resolve these mismatches. This PR addresses the issue #863.

### DIFF
--- a/beast/physicsmodel/dust/extinction_extension.py
+++ b/beast/physicsmodel/dust/extinction_extension.py
@@ -85,6 +85,9 @@ class F19_D03_extension(F19):
         ValueError
            Input x values outside of defined range
         """
+        # Ensure x has units; if it's a raw array, default it to Angstroms
+        if not hasattr(x, "unit"):
+            x = x * u.Angstrom
         # If it's already inverse microns, use it directly; otherwise, convert it
         if x.unit == u.micron**-1:
             x_um_inv = x
@@ -206,6 +209,9 @@ class G03_SMCBar_WD01_extension(G03_SMCBar):
         ValueError
            Input x values outside of defined range
         """
+        # Ensure x has units; if it's a raw array, default it to Angstroms
+        if not hasattr(x, "unit"):
+            x = x * u.Angstrom
         # If it's already inverse microns, use it directly; otherwise, convert it
         if x.unit == u.micron**-1:
             x_um_inv = x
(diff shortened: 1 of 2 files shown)


karllark:
There is an error that causes the code not to run.  Needs a fix of course.

galaxyumi:
@karllark, I've tried a couple of different approaches to address the git test errors. Since I can't replicate them in my local runs, I'm out of ideas—unless I modify the testing code inside the external `dust_extinction` package itself, which obviously isn't a good idea. Please let me know if you have any suggestions!

Just in case, my local python version is 3.12.3, Astropy version is 7.2, and Numpy version is 2.4.6. The different dependency versions might be a culprit.

karllark:
I wonder if you have an old version of dust_extinction installed.  Maybe try updating it and seeing if that changes things.

beast/physicsmodel/dust/extinction_extension.py
```diff
@@ -84,6 +85,15 @@ def evaluate(self, x, Rv):
...
            Input x values outside of defined range
         """
+        # Ensure x has units; if it's a raw array, default it to inverse microns
+        if not hasattr(x, "unit"):
+            x = x * u.micron**-1
+        # If it's already inverse microns, use it directly; otherwise, convert it
+        if x.unit == u.micron**-1:
+            x_um_inv = x
+        elif x.unit == u.Angstrom:
+            x_um_inv = x.to(u.micron**-1, equivalencies=u.spectral())
```
karllark: I think this can be just one line.  Line 95 should give both behaviors.  If already in 1/micron, it will do nothing.

beast/physicsmodel/dust/extinction_extension.py
```diff
@@ -199,22 +209,31 @@ def evaluate(self, x):
...
```
karllark: Same comment as above.

karllark (changes requested):
Looks great.  One minor comment that may reduce the number of lines of code.

beast/physicsmodel/dust/extinction_extension.py
```diff
@@ -72,7 +73,7 @@ def evaluate(self, x, Rv):
            expects either x in units of wavelengths or frequency
            or assumes wavelengths in wavenumbers [1/micron]
 
-           internally wavenumbers are used
+           internally wavenubmers in inverse micron are used
```
karllark: Argh.  Spelling.  Should be "wavenumbers."

karllark (approved):
One last spelling error to fix than then I'll merge.

karllark (approved):
Great!

PR #869, opened 2026-08-27
Accept explicit distance grids in `make_spectral_grid`

### LINKED ISSUE #868: `make_spectral_grid` cannot take an explicit (non-uniform) distance grid
### Description

`distance` in `make_spectral_grid` currently accepts only a single value or
`[min, max, step]`, which always produces a *uniformly* spaced grid:

```python
elif len(_distance) == 3:
    mindist, maxdist, stepdist = _distance
    distances = np.arange(mindist, maxdist + stepdist, stepdist)
elif len(_distance) == 1:
    distances = np.array(_distance)
else:
    raise ValueError("distance needs to be (min, max, step) or single number")
```

I'd love for there to be a way to supply a distance grid whose spacing varies (planning ahead for upcoming N66 and N83/N84 distance fitting in the SMC, hint hint)...

### Proposed fix

Let a list of more than three values be taken as the explicit grid. Lists of
one and three values keep their current meaning, so nothing existing changes.

christinawlindberg:
Closes #868

### What this does

Allows `distance` to be given as an explicit list of more than three values,
which is then used directly (sorted) as the distance grid. This enables
non-uniform, variable-resolution distance sampling — for example a grid whose
density follows the expected line-of-sight distribution of a galaxy.

### Backward compatibility

Unchanged for every currently valid input:

| input | behaviour |
|---|---|
| `60000` (scalar) | single distance, as before |
| `[min, max, step]` (3 values) | uniform `np.arange` grid, as before |
| `>3 values` | **new** — used as the explicit grid, sorted |
| `2 values` | `ValueError`, as before (message expanded) |

A three-element list is still always read as `[min, max, step]`, so an explicit
grid of exactly three distances cannot be expressed. That ambiguity is inherent
in the existing API; it is called out in the docstring and the docs rather than
resolved, to avoid changing what existing settings files mean.

### Also included

- `make_spectral_grid` docstring updated for the new form
- `docs/beast_setup.rst` entry for `distances` updated to match

### Context

Came out of building a shared physics-model grid for two SMC fields, where a
Gaussian-density distance grid gave the same effective resolution in 47 planes
instead of 66.

### DIFF
--- a/beast/physicsmodel/model_grid.py
+++ b/beast/physicsmodel/model_grid.py
@@ -226,14 +226,24 @@ def make_spectral_grid(
 
         # Construct the distances array. Turn single value into
         # 1-element list if single distance is given.
+        # A list of >3 values is used as an explicit (possibly non-uniform)
+        # distance grid, e.g. for variable-resolution distance sampling.
+        # NOTE: an explicit grid of exactly 3 values cannot be expressed
+        # (3 values are always interpreted as min, max, step) -- pad with a
+        # duplicate-free 4th value in that case.
         _distance = np.atleast_1d(distance)
         if len(_distance) == 3:
             mindist, maxdist, stepdist = _distance
             distances = np.arange(mindist, maxdist + stepdist, stepdist)
         elif len(_distance) == 1:
             distances = np.array(_distance)
+        elif len(_distance) > 3:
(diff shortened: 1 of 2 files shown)


karllark:
Can you please update the docs as appropriate for this change?

